Issue #6, opened 2018-09-20
K8s Ingress Controller


armab:
> Moved from https://github.com/StackStorm/k8s-st2/issues/31

As part of templating Helm Charts, we'll need a K8s Ingress controller (https://kubernetes.io/docs/concepts/services-networking/ingress/) to allow user configuring inbound access to StackStorm Web UI/APIs from the outside world and have more granular configuration comparing to `LoadBalancer` type.

## The Plan

Per https://github.com/StackStorm/stackstorm-ha/pull/44#issuecomment-459060262 discussion we want to:
1) Expose Ingress controller settings via Helm `values.yaml` to allow users to configure the SSL/TLS negotiation layer on their own (optional).
2) Change `st2web` Docker image so it will respond on HTTP by default (currently HTTPS).

https://github.com/StackStorm/stackstorm-enterprise-ha/blob/d5e34fc28e820e39097d5eaac7dcadcc4a481f2b/values.yaml#L204-L205

https://github.com/StackStorm/stackstorm-enterprise-ha/blob/d5e34fc28e820e39097d5eaac7dcadcc4a481f2b/templates/ingress.yaml#L1

## Resources
- https://kubernetes.io/docs/concepts/services-networking/ingress/
- https://github.com/kubernetes/ingress-nginx
- https://kubernetes.github.io/ingress-nginx/ingress-controller-catalog/
- https://blog.getambassador.io/kubernetes-ingress-nodeport-load-balancers-and-ingress-controllers-6e29f1c44f2d
- https://akomljen.com/kubernetes-nginx-ingress-controller/

armab:
Overview of different types K8s Service types and when to use what:
https://medium.com/@pczarkowski/kubernetes-services-exposed-86d45c994521

mosn:
Providing my feedback since this one is still open.
with changes to `values.yaml` it was easy to expose service behind `st2web` to outside.

```
  service:
    type: "LoadBalancer"
    hostname: "example.domain.com"
```

I can also wish for a way to set Ingress controller inside `values.yaml`  but I am not sure how feasible that is considering all services depending on st2web load balancing capability.  

```
ingress:
  enabled: true
  path: /
  annotations:
    kubernetes.io/ingress.class: nginx
    certmanager.k8s.io/issuer:  my-issuer
    nginx.ingress.kubernetes.io/proxy-body-size: 50m
    nginx.ingress.kubernetes.io/proxy-send-timeout: "600"
[...]

PR #34, opened 2018-11-07
Add LoadBalancer type to st2web service

warrenvw:
As part of the journey toward allowing user to use an ingress controller (#6), this PR allows user to choose between using the simpler `NodePort` or `LoadBalancer` for the st2web service.


requirements.yaml
```diff
@@ -14,3 +14,6 @@ dependencies:
     version: 0.2.2
     repository: https://kubernetes-charts-incubator.storage.googleapis.com/
     condition: docker-registry.enabled
+  - name: external-dns
```
armab: Because there could be `NodePort` type, this dependency needs a condition to download it only if service type is LoadBalancer

warrenvw: Yes. Thanks.

values.yaml
```diff
@@ -212,8 +212,13 @@ st2web:
       cpu: "50m"
     limits:
       memory: "100Mi"
-  # TODO: Add Ingress and/or LoadBalancer setting as a way to expose service to public (#6)
+  # TODO: Add Ingress setting as a way to expose service to public (#6).
+  # For now, use LoadBalancer or NodePort (default to LoadBalancer).
   # ingress:
+  service:
+    # NodePort | LoadBalancer
```
armab: There could be also `ClusterIP` and `ExternalName`

values.yaml
```diff
@@ -212,8 +212,13 @@ st2web:
       cpu: "50m"
     limits:
       memory: "100Mi"
-  # TODO: Add Ingress and/or LoadBalancer setting as a way to expose service to public (#6)
+  # TODO: Add Ingress setting as a way to expose service to public (#6).
+  # For now, use LoadBalancer or NodePort (default to LoadBalancer).
   # ingress:
+  service:
```
armab: `service` will need to expose in values.yaml optional `annotation` field to configure additional settings is it SSL or AWS-load-balancer specific or something else

See an example: https://kubernetes.io/docs/concepts/services-networking/service/#ssl-support-on-aws

values.yaml
```diff
@@ -212,8 +212,13 @@ st2web:
...
+    # NodePort | LoadBalancer
+    type: "LoadBalancer"
```
armab: Not 100% sure, but maybe `NodePort` or `ClusterIP` (default K8s) would be a better default?
With an additional option to set it to `LoadBalancer` or other type.

armab:
With adding more options for service types, need to handle it also correctly in Helm chart `NOTES.txt` so it'll show the st2web IP/URL correctly depending on resource.

See example: https://github.com/helm/charts/blob/master/stable/drupal/templates/NOTES.txt#L6-L37

templates/services.yaml
```diff
@@ -80,6 +80,24 @@ metadata:
   name: {{ .Release.Name }}-st2web{{ template "enterpriseSuffix" . }}
   annotations:
     description: StackStorm st2web, - an admin Web UI and main entry point for external API requests
+    {{- if .Values.st2web.service.hostname }}
+    external-dns.alpha.kubernetes.io/hostname: {{ .Values.st2web.service.hostname | quote }}
+    {{- end }}
+    # For more information, see https://kubernetes.io/docs/concepts/services-networking/service/#ssl-support-on-aws
+    {{- if eq .Values.st2web.service.type "LoadBalancer" }}
+    {{- if .Values.st2web.service.loadBalancer.sslCert }}
+    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: {{ .Values.st2web.service.loadBalancer.sslCert | quote }}
```
armab: That's not exactly what user will be looking. The problem is that there could be _much_ more possible settings to configure via annotations which we can't template and cover.

Just expose `annotations` in values.yaml and let them decide what's needed.
This is a common practice, according to Helm charts repo.

Something like https://github.com/helm/charts/blob/master/stable/cockroachdb/values.yaml#L47-L49

templates/NOTES.txt
```diff
@@ -8,9 +8,13 @@ Congratulations! You have just deployed StackStorm {{ if .Values.enterprise.enab
   ╚══════╝   ╚═╝   ╚══════╝    ╚═╝  ╚═╝╚═╝  ╚═╝     ╚═════╝ ╚═╝  ╚═╝
 
 1. Get the StackStorm Web UI URL by running these commands:
+{{- if ne .Values.st2web.service.type "LoadBalancer" }}
```
armab: Need more cases for other Service types, as we now open this setting in values.yaml and expect that `NOTES.txt` will show service endpoint correctly.

values.yaml
```diff
@@ -212,8 +212,18 @@ st2web:
       cpu: "50m"
     limits:
       memory: "100Mi"
-  # TODO: Add Ingress and/or LoadBalancer setting as a way to expose service to public (#6)
+  # TODO: Add Ingress setting as a way to expose service to public (#6).
+  # For now, use LoadBalancer or NodePort (default to LoadBalancer).
```
armab: With recent changes, default is not `LoadBalancer` anymore

templates/NOTES.txt
```diff
@@ -8,9 +8,13 @@ Congratulations! You have just deployed StackStorm {{ if .Values.enterprise.enab
...
 export ST2WEB_IP=$(minikube ip 2>/dev/null || kubectl get nodes --namespace {{ .Release.Namespace }} -o jsonpath="{.items[0].status.addresses[0].address}")
-export ST2WEB_PORT=$(kubectl get --namespace {{ .Release.Namespace }} -o jsonpath="{.spec.ports[0].nodePort}" services {{ .Release.Name }}-st2web{{ template "enterpriseSuffix" . }})
-echo https://${ST2WEB_IP}:${ST2WEB_PORT}/
+export ST2WEB_PORT=":$(kubectl get --namespace {{ .Release.Namespace }} -o jsonpath="{.spec.ports[0].nodePort}" services {{ .Release.Name }}-st2web{{ template "enterpriseSuffix" . }})"
```
armab: `:` definitely should not be part of the `PORT` :)

values.yaml
```diff
@@ -212,8 +212,21 @@ st2web:
...
       memory: "100Mi"
-  # TODO: Add Ingress and/or LoadBalancer setting as a way to expose service to public (#6)
-  # ingress:
+  # TODO: Add Ingress setting as a way to expose service to public (#6).
+  ingress: {}
+  #  hostname:
+  service:
+    # type can be one of "ClusterIP", "NodePort", "LoadBalancer" or "ExternalName"
+    type: "NodePort"
+    hostname: ""
```
armab: Every record in values.yaml ideally to have a description comment

values.yaml
```diff
@@ -212,8 +212,21 @@ st2web:
       cpu: "50m"
     limits:
       memory: "100Mi"
-  # TODO: Add Ingress and/or LoadBalancer setting as a way to expose service to public (#6)
-  # ingress:
+  # TODO: Add Ingress setting as a way to expose service to public (#6).
+  ingress: {}
+  #  hostname:
```
armab: This looks confusing because we're not adding yet an `Ingress` controller as K8s resource.
Especially keeping in mind `service.hostname`.

values.yaml
```diff
@@ -212,8 +212,16 @@ st2web:
...
+  # TODO: Add Ingress setting as a way to expose service to public (#6).
   # ingress:
+  service:
+    # type can be one of "ClusterIP", "NodePort", "LoadBalancer" or "ExternalName"
+    type: "NodePort"
+    # The hostname associated with st2web service (externalName, added to external DNS, etc.)
+    hostname: ""
+    annotations: {}
+    # Example annotations. For more information, see
+    # https://kubernetes.io/docs/concepts/services-networking/service/#ssl-support-on-aws
```
armab: Need to move comments before the `annotations` definition.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## In Development
-
+* Add ability to specify service type for st2web (#35)
```
armab: Probably need to add a new version `v0.8.0` here too, as it was bumped in Helm metadata already

templates/NOTES.txt
```diff
@@ -7,11 +7,29 @@ Congratulations! You have just deployed StackStorm {{ if .Values.enterprise.enab
...
+NOTE: It may take a few minutes for the LoadBalancer IP to be available.
+      Watch the status with "kubectl get service st2cicd-st2web-enterprise -o jsonpath='{.status.loadBalancer.ingress[0].hostname}'"
+
+export ST2WEB_IP=$(kubectl get service st2cicd-st2web-enterprise -o jsonpath='{.status.loadBalancer.ingress[0].hostname}')
+echo https://${ST2WEB_IP}/
+
+{{- else if contains "ClusterIP" .Values.st2web.service.type }}
+
+echo https://127.0.0.1:8080
+kubectl port-forward --namespace {{ .Release.Namespace }} svc/{{ .Release.Name }}-st2web-{{ if .Values.enterprise.enabled }}-enterprise {{ end }} 8080:80
```
armab: Need `443` port here really, not `80` as that's default for st2web

templates/NOTES.txt
```diff
@@ -7,11 +7,29 @@ Congratulations! You have just deployed StackStorm {{ if .Values.enterprise.enab
...
+
+NOTE: It may take a few minutes for the LoadBalancer IP to be available.
+      Watch the status with "kubectl get service st2cicd-st2web-enterprise -o jsonpath='{.status.loadBalancer.ingress[0].hostname}'"
+
+export ST2WEB_IP=$(kubectl get service st2cicd-st2web-enterprise -o jsonpath='{.status.loadBalancer.ingress[0].hostname}')
+echo https://${ST2WEB_IP}/
+
+{{- else if contains "ClusterIP" .Values.st2web.service.type }}
+
+echo https://127.0.0.1:8080
```
armab: Then it will be something like `8443` here, instead of `8080`

PR #43, opened 2019-01-10
Add link to st2web configuration

trstruth:
The docs mention that some additional configuration may be required to expose st2webui and the services it proxifies, but is rather vague about how that may be done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,8 @@ kubectl exec -it ${ST2CLIENT} /bin/bash
 ### [st2web](https://docs.stackstorm.com/latest/reference/ha.html#nginx-and-load-balancing)
 st2web is a StackStorm Web UI admin dashboard. By default, st2web K8s config includes a Pod Deployment and a Service.
 `2` replicas (configurable) of st2web serve the web app and proxy requests to st2auth, st2api, st2stream.
-> **Note!** K8s Service uses only NodePort at the moment, so installing this chart will not provision a K8s resource of type LoadBalancer or Ingress (TODO!).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,7 +67,7 @@ kubectl exec -it ${ST2CLIENT} /bin/bash
 st2web is a StackStorm Web UI admin dashboard. By default, st2web K8s config includes a Pod Deployment and a Service.
 `2` replicas (configurable) of st2web serve the web app and proxy requests to st2auth, st2api, st2stream.
 > **Note!** K8s Service uses only NodePort at the moment, so installing this chart will not provision a K8s resource of type LoadBalancer or Ingress (TODO!).
```
armab: We need to rephrase this note as we do allow to configure LoadBalncer as service type for st2web.
Seems we forgot to do that.

README.md
```diff
@@ -67,7 +67,7 @@ kubectl exec -it ${ST2CLIENT} /bin/bash
...
-  Depending on your Kubernetes cluster setup you may need to add additional configuration to access the Web UI service or expose it to public net.
+  Depending on your Kubernetes cluster setup you may need to add additional [configuration](https://github.com/StackStorm/stackstorm-ha/blob/master/values.yaml#L218-L219) to access the Web UI service or expose it to public net.
```
armab: I’d prefer to note explicitly a mention that st2web K8s service type could be configured via Helm values to expose the ST2 APIs and Web UI.
That would be better comparing to including a link.

armab:
Additionally, after merging this, we’ll need to update respective st2docs section for K8s st2web 
https://docs.stackstorm.com/install/k8s_ha.html#st2web

README.md
trstruth: I agree, the line number reference could be easily broken.  I've updated it.

warrenvw:
Agree it looks better. Thanks guys for doing this!

warrenvw:
I'll update the st2docs.

Issue #16, opened 2018-09-20
Autogenerate secrets if no Helm value provided, remove insecure defaults


armab:
In Helm values we specify unsafe defaults for ~SSL auto-generated cert #6~, SSH key, st2 password to bring a "one click" experience.

An alternative to that is to auto-generate these secrets if no Helm values provided.

armab:
Surprisingly, helm via Spring templating engine has a lot of crypto functions: https://masterminds.github.io/sprig/crypto.html
`genPrivateKey` and `genSelfSignedCert` is all we need.

So technically it's all possible, just a matter of implementation.

armab:
As described in https://github.com/StackStorm/stackstorm-ha/pull/62, Helm doesn't have enough functionality to transition from autogenerated vs custom values-defined state and so SSH key in Helm values remains as is until better instrumentation from Helm introduced.

We however can get rid of st2web/nginx SSL certificates from Helm values switching to HTTP by default and allow user to configure HTTPS as optional Ingress controller on top of st2web, - the plan described in #6

armab:
Per https://github.com/helm/helm/issues/3053#issuecomment-628876519 looks like Helm3 has a new solution via new `lookup` feature.
We'll need to review this.

PR #44, opened 2019-01-29
Add support to specify images name and to enable HTTP for st2web

GGabriele:
Hello, this is an effort to add the possibility to specify the images name to pull from a custom registry. If no image name is specified, the default behavior is used.

I'm also enabling the toggling of HTTP/HTTPS for st2web (another PR to patch the st2web container will follow).

Let me know if any of this doesn't make sense :)

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -204,6 +204,8 @@ secrets:
 ##
 # Many st2web instances, placed behind a load balancer that serve web app and proxify requests to st2auth, st2api, st2stream.
 st2web:
+  protocol:
+    http: false
   # Minimum 2 replicas are recommended to run st2web in HA mode
   replicas: 2
   # Tested resource consumption based on multiple requests to st2web within nginx
--- a/values.yaml
+++ b/values.yaml
@@ -100,6 +100,9 @@ st2:
 # TODO: Move to `secrets.yaml` when it gets implemented in Helm (https://github.com/kubernetes/helm/issues/2196) ? (#14)
 # TODO: Alternatively as part of reorganizing Helm values, consider moving values to existing `st2` and `st2web` sections ? (#14)
 secrets:
+  images:
+    # Pull images from a custom registry
+    customRegistry:
   st2:
     # Username, used to login to StackStorm system
     username: st2admin
@@ -196,7 +199,6 @@ secrets:
       ofUOhwAXuMsTtuLagOMyAJVs+KRVrvnXGT/p9l213ZAnDtFSpkvcjD9WUcupeTca
       BjdoJBzImjVB5znOgIui3ME5
       -----END PRIVATE KEY-----
-
 ##
 ## StackStorm HA Cluster pod settings for each individual service/component.
 ##
@@ -226,6 +228,8 @@ st2web:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  image:
+    imageName:
 # https://docs.stackstorm.com/reference/ha.html#st2auth
 # Multiple st2auth processes can be behind a load balancer in an active-active configuration.
 st2auth:
@@ -236,6 +240,8 @@ st2auth:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  image:
+    imageName:
 # https://docs.stackstorm.com/reference/ha.html#st2api
 # Multiple st2api process can be behind a load balancer in an active-active configuration.
 st2api:
@@ -246,6 +252,8 @@ st2api:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  image:
+    imageName:
 # https://docs.stackstorm.com/reference/ha.html#st2stream
 # Multiple st2stream process can be behind a load balancer in an active-active configuration.
 st2stream:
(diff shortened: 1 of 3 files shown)


armab:
We tried to keep the `values.yaml` less overwhelmed with the configuration, awaiting for contributions like this to appear when there is a need from our users.
Good to see this happening! :+1: 

One question. Do you really need to change Docker image name for each service or specifying a custom registry would be enough? If you could keep the image names the same with original st2 images, then you won't need any custom names. Would it solve your case? Let me know your thoughts.

values.yaml
```diff
@@ -196,12 +199,13 @@ secrets:
...
       BjdoJBzImjVB5znOgIui3ME5
       -----END PRIVATE KEY-----
-
 ##
 ## StackStorm HA Cluster pod settings for each individual service/component.
 ##
 # Many st2web instances, placed behind a load balancer that serve web app and proxify requests to st2auth, st2api, st2stream.
 st2web:
+  protocol:
+    http: false
```
armab: As for `st2web` HTTP/HTTPS.

I agree that the feature makes sense too, but let's please split it into a separated PR as that's a different code and discussion.

A quick story there: we thought about exposing `st2web` via HTTP by default and offload HTTPS/certificates to optional Service Type like Ingress Controller #6 or LoadBalancer settings, but not sure, might need more input.
For reference, how other Helm roles usually do this: https://github.com/helm/charts/tree/master/stable/kibana#configuration https://github.com/helm/charts/tree/master/stable/hubot#configuration

GGabriele:
Hi @armab , thanks for reviewing this.

> One question. Do you really need to change Docker image name for each service or specifying a custom registry would be enough?

Ideally I don't want to build all the images, but just few of them (in this case, only st2web actually). Maybe we can name the images following the original st2 images, but we'd still need a way to specify if an image should be pulled from the custom registry or not, e.g.:

```
st2web:
  images:
    custom: true
```

That said, I wouldn't even need to build/pull any custom image if we can solve the HTTP/HTTPS problem in the upstream, so let me know if I can help tackling it somehow :)

armab:
We talked with @warrenvw and how that can fit the roadmap/plans.

In short, in future we indeed want to expose st2web in K8s via HTTP by default.
On top of that, allow users to configure HTTPS/SSL settings via K8s Ingress controller (not yet done https://github.com/StackStorm/stackstorm-ha/issues/6) or LoadBalancer annotations (done).

This way SSL termination responsibility will shift from `st2web` nginx to K8s-controlled optional layers. It looks like a obvious practice across https://github.com/helm/charts. This way we'll also get rid of beta dirt like https://github.com/StackStorm/stackstorm-ha/blob/master/values.yaml#L140-L198 which will improve https://github.com/StackStorm/stackstorm-ha/issues/16 story.

First part (st2web HTTP) will require adjusting https://github.com/StackStorm/st2-dockerfiles/tree/master/st2web and K8s repo to reflect the changes.

Second part is adding Ingress controller https://github.com/StackStorm/stackstorm-ha/issues/6 option to `stackstorm-ha` to compensate the absense of HTTPS by default.

----

@GGabriele How that fits your story, at what point of your infra you plan to add secure layer on top of HTTP-only st2web?

GGabriele:
This makes sense and it sounds great!
In my case, what I did is adjust the [st2web](https://github.com/StackStorm/st2-dockerfiles/tree/master/st2web) container to support HTTP and use it while adding a security layer directly in GCP with an [IAP](https://cloud.google.com/iap/docs/), so that I can authenticate using emails account.

armab:
@GGabriele Perfect! Want to contribute the changes then?

For https://github.com/StackStorm/st2-dockerfiles/tree/master/st2web I think we'll need a switch so nginx will pick up HTTP or HTTPS configs on startup based on user's specified ENV variable. This will keep some level of compatibility for those who need HTTPS from that Docker image.

cognifloyd (changes requested):
Could you rebase this PR or merge in master?

Several things have changed, including at least:
- the enterprise stuff is gone, so maybe that would simplify some of these `imageName` blocks.
- the st2web pod is using HTTP now instead of HTTPS. I'm not sure when that changed, but if an HTTP/HTTPS flag is still needed, maybe that could be put in a separate PR?
- looks like we can get the custom registry bit from `image.repository` now, right?

cognifloyd:
#200 was just merged which adds the ability to define a custom tag for each of the pods. That will also affect making image names configurable in this PR.

armab:
@cognifloyd Yeah indeed, good call :+1:  From the conversation history:

1)
> For https://github.com/StackStorm/st2-dockerfiles/tree/master/st2web I think we'll need a switch so nginx will pick up HTTP or HTTPS configs on startup based on user's specified ENV variable. This will keep some level of compatibility for those who need HTTPS from that Docker image.

2)
> In short, in future we indeed want to expose st2web in K8s via HTTP by default.
On top of that, allow users to configure HTTPS/SSL settings via K8s Ingress controller (not yet done #6) or LoadBalancer annotations (done).
>
>This way SSL termination responsibility will shift from st2web nginx to K8s-controlled optional layers. It looks like a obvious practice across https://github.com/helm/charts. This way we'll also get rid of beta dirt like https://github.com/StackStorm/stackstorm-ha/blob/master/values.yaml#L140-L198 which will improve #16 story.
>
> First part (st2web HTTP) will require adjusting https://github.com/StackStorm/st2-dockerfiles/tree/master/st2web and K8s repo to reflect the changes.
>
> Second part is adding Ingress controller #6 option to stackstorm-ha to compensate the absense of HTTPS by default.

Both features were implemented a while ago.
The st2web listens on HTTP by default and SSL termination happens and configurable in another layer: Ingress controller or LoadBalancer.

@GGabriele Thanks for the initial PR, the proposed approach and discussion here helped to eventually implement it the right way!

PR #45, opened 2019-02-05
Add new Helm value setting 'st2.apikeys' to import ST2 API keys

### LINKED ISSUE #36: Add support for 'st2 apikey load'
Similar as we do with importing data into st2 keyvalue storage #30, need an option to import st2 apikeys.
Respective documentation: https://docs.stackstorm.com/authentication.html#api-key-migration

As needed in https://github.com/StackStorm/ops-infra/blob/master/roles/stackstorm.st2cicd/tasks/install_api_keys.yml

armab:
Closes #36 

Implements https://docs.stackstorm.com/authentication.html#api-key-migration mechanism to import previously exported list of ST2 API keys that's needed for reproducible deployments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## In Development
 
+## v0.9.0
+* Add new Helm value setting `st2.apikeys` to allow importing predefined ST2 API keys (#36)
+
 ## v0.8.4
 * Pin st2 version to `v3.0dev` as a new latest development version (#41)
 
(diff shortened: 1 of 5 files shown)


armab:
The feature in K8s/HA is ready, however it's blocked by ST2 regression when `st2 apikey load` doesn't work in an idempotent way and can't import already imported API keys:
```
ERROR: 409 Client Error: Conflict
MESSAGE: Tried to save duplicate unique keys (E11000 duplicate key error collection: st2.api_key_d_b index: key_hash_1 dup key: { : "56928c2d9637ce44338e9564d4b939df8b258 ...." }) for url: http://127.0.0.1:9101/v1/apikeys
```

Similar bug was fixed before, but it reappeared for some new reason.
I'm working on fixing ST2 next.

armab:
Can now confirm the fixed behavior and working Helm values functionality after getting new Docker images based on latest `st2` with https://github.com/StackStorm/st2/pull/4542

PR #50, opened 2019-02-23
Bump dependency versions

warrenvw:
Use latest versions of the `rabbitmq-ha` and `mongodb-replicaset` charts.


warrenvw:
I've bumped all other dependencies, updated chart version and added an entry in the changelog.

armab:
Per previous discussion, if new rabbitmq-ha Helm chart includes pod naming changes, then it needs more work TBD to verify if ST2 HA Helm templating works with those upstream changes well.

Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 # Update StackStorm version here to rely on other Docker images tags
 appVersion: 3.0dev
 name: stackstorm-ha
-version: 0.9.0
+version: 0.10.0
```
armab: Ideally version needs to be bumped just before the release, as by the fact this is not yet `0.10.0`.

warrenvw:
Note that rabbitmq-ha `1.21.0` has a problem with the security settings / permissions, so I held its version at `1.20.1`.

Chart.yaml
warrenvw: Oops, this slipped in there.. agreed.

armab:
Useful info, It was a good opportunity to include that in code comments.

PR #51, opened 2019-03-01
Allow st2sensorcontainer to be partitioned

### LINKED ISSUE #4: Partitioning StackStorm Sensors
> Moved from https://github.com/StackStorm/k8s-st2/issues/12

See: https://docs.stackstorm.com/reference/sensor_partitioning.html

At the moment `st2sensorcontainer` K8s configuration consists of Deployment with hardcoded `1` replica and works as a parent responsible to forking/starting/stopping all sensors in the system.

Future plans to re-work this setup (discussion https://github.com/StackStorm/discussions/issues/305) and benefit from Docker-friendly [single-sensor-per-container mode](https://github.com/StackStorm/st2/pull/4179) (added since st2 `v2.9`) as a way of [Sensor Partitioning](https://docs.stackstorm.com/latest/reference/sensor_partitioning.html), distributing the computing load between many pods and relying on K8s failover/reschedule mechanisms, instead of running everything on 1 single instance of st2sensorcontainer.

Need to codify + config that mode in Docker + K8s Helm templates so user specifies list of sensors he wants to run in `values.yaml`, Helm takes care about starting a pod for each record.

### References

https://github.com/StackStorm/stackstorm-enterprise-ha/blob/ca116b1c20b8953aebde429139022bd4141aa068/values.yaml#L279-L285

https://github.com/StackStorm/stackstorm-enterprise-ha/blob/ca116b1c20b8953aebde429139022bd4141aa068/templates/deployments.yaml#L710-L715

warrenvw:
Resolves #4.

Benefit from Docker-friendly [single-sensor-per-container mode](https://github.com/StackStorm/st2/pull/4179) (added since st2 `v2.9`) as a way of [Sensor Partitioning](https://docs.stackstorm.com/latest/reference/sensor_partitioning.html), distributing the computing load between many pods and relying on K8s failover/reschedule mechanisms, instead of running everything on 1 single instance of st2sensorcontainer.

Remaining work:

- [x] Documentation in `README.md`
- [x] Documentation updates in st2docs repo after doc changes approved in this repo.


warrenvw:
I'm thinking about releasing a new chart that includes this PR and #50. Need to determine the appVersion (2.10dev or 3.0dev?)... and chart version (v0.10.0?).

armab:
I'm good with releasing it in one Helm chart version.

Need to make sure that PR are merged separately and tested in isolation, especially considering: https://github.com/StackStorm/stackstorm-ha/pull/50#issuecomment-468346661

`appVersion` remains the same as we're relying on `dev` and there is no `2.10dev`. See https://github.com/StackStorm/st2/blob/master/st2common/st2common/__init__.py

templates/deployments.yaml
```diff
@@ -843,8 +843,8 @@ spec:
             /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
       # System packs
       - name: st2-system-packs
-        image: "{{ template "imageRepository" . }}/st2actionrunner{{ template "enterpriseSuffix" . }}:{{ .Chart.AppVersion }}"
-        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        image: "{{ template "imageRepository" $ }}/st2actionrunner{{ template "enterpriseSuffix" $ }}:{{ $.Chart.AppVersion }}"
+        imagePullPolicy: {{ $.Values.image.pullPolicy }}
```
armab: Curious what is the trick with `.` vs `$.`?
I don't see this widely demonstrated in Helm examples or other charts. Does it fixes anything, what's the difference?

templates/deployments.yaml
```diff
@@ -780,56 +780,56 @@ spec:
 {{ toYaml . | indent 8 }}
     {{- end }}
 
+{{- range .Values.st2.packs.sensors }}
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ .Release.Name }}-st2sensorcontainer{{ template "enterpriseSuffix" . }}
+  name: {{ $.Release.Name }}-st2sensorcontainer-{{ .name }}{{ template "enterpriseSuffix" $ }}
```
armab: Just thinking out loud: it could be a matter of time when `name` exceeds the limit of number of allowed characters.
But we have no solutions here, except of cutting/trimming it. A task for some future.

values.yaml
```diff
@@ -85,6 +85,24 @@ st2:
       name: st2packs
       tag: latest
       pullPolicy: Always
+
+    # https://docs.stackstorm.com/reference/ha.html#st2sensorcontainer
+    # It is possible to run st2sensorcontainer in HA mode by running one process on each compute instance.
+    # Each sensor node needs to be provided with proper partition information to share work with other sensor
+    # nodes so that the same sensor does not run on different nodes.
+    sensors:
```
armab: Documentation/examples and references to it should be definitely improved.

Take a look at README.md. Additionally, we need to have a section there describing the full example, why is this needed and defaults.

armab:
BTW, did you figure out why the K8s jobs didn't trigger (as part of the previous iteration)?

armab:
TODO: once it's merged we'll need a st2docs sync-up.

templates/deployments.yaml
warrenvw: Ref: https://helm.sh/docs/chart_template_guide/#variables

> The `range` function will “range over” (iterate through) a list. But now something interesting happens. Just like with sets the scope of `.`, so does a `range` operator.
> 
> However, there is one variable that is always global - `$` - this variable will always point to the root context. This can be very useful when you are looping in a range and need to know the chart’s release name.

The scope of the range block is `.Values.st2.packs.sensors`. So any of the values outside that scope need to be referred to with `$`. I'm open to any other "cleaner" solution.

armab: Makes sense, thanks for explanation!

values.yaml
```diff
@@ -85,6 +85,24 @@ st2:
...
+      # Default "all-in-one" sensorcontainer that runs all sensors.
+      # To partition sensors with one sensor per node, override st2.packs.sensors.
+      # NOTE: Do not modify this file.
+      - name: aio
```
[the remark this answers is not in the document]
warrenvw: Not a bad idea.. i'll look into keeping the same previous name (without a postfix). On my first pass, the code was simpler if `.name` was used for all cases, but I'll see if I can come up with an elegant solution that does what you're asking.

values.yaml
warrenvw: 👍 I definitely need to add documentation to configure st2sensor partitioning. I'll use some examples to help clarify.

warrenvw:
I would imagine that the jobs didn't trigger due to some issue with Helm. I do not know the precise root cause at this point in time.

See https://github.com/stackstorm/discussions/issues/318#issuecomment-468190021 for more details. I didn't bother checking this earlier because I had no reason to think helm would be so poorly behaved. I'd always installed helm chart within the chart workspace. I didn't ever have to install from a packaged chart just to install correctly.

values.yaml
```diff
@@ -85,6 +85,24 @@ st2:
...
+        settings:
```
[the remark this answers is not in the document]
warrenvw: I'll experiment with this. I don't see any reason why not. I'd done it this way because one of the examples I'd seen grouped variables under `settings`. Didn't seem like a bad idea at the time.

templates/deployments.yaml
warrenvw: The current container names are about 75 characters long including the resource prefix `/api/v1/pods`. The maximum length of a Kubernetes resource is 253 characters. Unless I'm missing something obvious, I do not think we'll exceed any limits.

armab: See:
https://github.com/helm/charts/blob/d712e0cfd7d1426b6adb59473530fcc84bacbaf2/stable/rabbitmq-ha/templates/_helpers.tpl#L11-L12
https://github.com/helm/helm/issues/2022

templates/deployments.yaml
```diff
@@ -856,23 +856,32 @@ spec:
...
       containers:
-      - name: st2sensorcontainer{{ template "enterpriseSuffix" . }}
-        image: "{{ template "imageRepository" . }}/st2sensorcontainer{{ template "enterpriseSuffix" . }}:{{ .Chart.AppVersion }}"
-        imagePullPolicy: {{ .Values.image.pullPolicy }}
+      - name: st2sensorcontainer-{{ .name }}{{ template "enterpriseSuffix" $ }}
+        image: "{{ template "imageRepository" $ }}/st2sensorcontainer{{ template "enterpriseSuffix" $ }}:{{ $.Chart.AppVersion }}"
+        imagePullPolicy: {{ $.Values.image.pullPolicy }}
         # TODO: Add liveness/readiness probes (#3)
         #livenessProbe:
         #readinessProbe:
```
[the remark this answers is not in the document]
warrenvw: Linking issue #3, as this PR helps resolve some of that issue.

README.md
```diff
@@ -112,11 +112,44 @@ In an HA deployment there must be a minimum of `2` replicas of st2notifier runni
...
-between many pods and relying on K8s failover/reschedule mechanisms, instead of running everything on `1` single instance of st2sensorcontainer.
+st2sensorcontainer manages StackStorm sensors: It starts, stops and restarts them as subprocesses.
+By default, deployment is configured with `1` replica containing all the sensors.
+
+st2sensorcontainer also supports a more Docker-friendly single-sensor-per-container mode as a way of
+[Sensor Partitioning](https://docs.stackstorm.com/latest/reference/sensor_partitioning.html). This
+distributes the computing load between many pods and relies on K8s failover/reschedule mechanisms,
+instead of running everything on a single instance of st2sensorcontainer. To partition the sensors,
+create a yaml file containing `st2.packs.sensors`, and at a minimum, the `name` and `ref` elements.
+You can also specify a `livenessProbe` and `readinessProbe` that Kubernetes will use to check
```
armab: This part about the probes, resources and rest could be omitted to keep documentation balanced.

README.md
```diff
@@ -112,11 +112,44 @@ In an HA deployment there must be a minimum of `2` replicas of st2notifier runni
...
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        resources: {}
+        nodeSelector: {}
+        affinity: {}
+        tolerations: {}
+```
+
+Pass the name of this file to `helm install` using the `-f <file>` option. Add additional sensors to
+the `sensors:` list following the same format as above.
```
armab: I think it worth also mentioning that particular sensor should be shipped as part of the Dockerized packs.

PR #52, opened 2019-03-01
HA for etcd

### LINKED ISSUE #8: HA for etcd cluster
> Moved from https://github.com/StackStorm/st2enterprise-dockerfiles/issues/48

Currently we install single-node `etcd` instance as st2 `[coordination]` backend, just to get started quick.

Enhance the K8s manifests to configure `3` node HA cluster by default.

In future, further template with Helm to optionally use 5, 7 or more nodes, if user needs it.

Use 3rd party Helm chart for configuring `etcd` similar to `mongodb-ha` and `rabbitmq-ha`, instead of using quick hacked K8s objects.

See https://github.com/helm/charts/tree/master/incubator/etcd

armab:
Closes #8 

Use `etcd` the right way, HA-friendly as an external Helm dependency.
Replace single-node `etcd`coordination backend with 3-node etcd cluster via https://github.com/helm/charts/tree/master/incubator/etcd


armab:
TODO: st2docs update

PR #55, opened 2019-03-15
Chatops integration

### LINKED ISSUE #17: Add st2chatops
Currently K8s/Helm is missing st2chatops deployment.

Making it Docker and K8s-friendly to read ENV vars outside of `st2chatops.env` will take some work, (see blocker https://github.com/StackStorm/st2chatops/issues/50).

And so running 1 replica of st2chatops would be doable and nice to have, but having st2chatops in `HA` is a story for a new dedicated issue and research.

rapittdev:
Closes #17 

Proposal for chatops integration base on #17 

As mentioned in the tracking issue:
- [ ] Update `st2docs`
- [x]  start with defining a Dockerfile for st2chatops 
- [x]  st2chatops configuration could be controlled by passing ENV variables to override defaults
- [x]  creating a K8s resource for st2chatops as well as corresponding Helm values

find out more about the why [here](https://mattermost.com/blog/rapitt-mattermost-quick-messaging-open-source-extensibility/)


mosn:
Thank you @armab for the suggestions. I ended up doing that after tested the other approaches. as suggested, This is a base for discussions/suggestions. 

I took this repository as a stepping stone for the new integration, There aren't much of division form it; However, there might be unnecessary values passed around which is left for cleaning up. 

Corresponding  [st2-dockerfile PR ](https://github.com/StackStorm/st2-dockerfiles)  In my approach I am using st2chatops package. TBD if there's a better source.

I probably need a few more days to bring it to a clean state, feedback, contributions very welcome as I won't be able to commit time next week.

mosn:
Hey Sure, 

This branch is outdated as I worked on it locally. Will push changes later today. 

Thanks

values.yaml
```diff
@@ -85,41 +86,66 @@ st2:
...
+      ip: 
+      groups:
+      user:
+      password:
+      wss_port: 443
+      http_port: 443
+      reply: true
+      log_level: info
+      reply: true
+    mattermost.yaml: |
```
armab: Instead of having dedicated section for each possible hubot adapter (which could be outdated with time), would be better for user to specify any custom ENV variables they want to pass to `st2chatops` container.

Those custom ENVs will be applied on top of `st2chatops.env` like https://github.com/StackStorm/st2chatops/blob/c99d779d6e67385156bfd11b4c9997cf7e6f7ca8/st2chatops.env#L25-L26 in this example.

values.yaml
```diff
@@ -85,41 +86,66 @@ st2:
...
-  apikeys:
-    #- created_at: '2018-12-15T00:21:48.507388Z'
-    #  enabled: true
-    #  id: 5c14491c6cb8de1a9207e3a2
-    #  key_hash: 56928c2d9637ce44338e9564d4b939df8b258410db23b5a80f8ad69d58e648b574f35f9293c3a76bde263738be9aa8379a81553cd55513ad672540b7b0ec0cac
-    #  metadata: {"comment": "Example unsecure ST2 API key from K8s HA Helm values.yaml"}
-    #  uid: api_key:56928c2d9637ce44338e9564d4b939df8b258410db23b5a80f8ad69d58e648b574f35f9293c3a76bde263738be9aa8379a81553cd55513ad672540b7b0ec0cac
-    #  user: st2admin
+  # Experimental st2chatops integration for StackStorm-HA Helm chart.  (https://github.com/mosn/stackstorm-ha/issues/17)
+  chatops:
```
armab: Let's please move chatops settings into existing `st2chatops` section you defined later.

values.yaml
```diff
@@ -85,41 +86,66 @@ st2:
...
+    gloal.yaml: |
+      ---
+      image: "stackstorm/st2chatops"
+      reoisitory:
```
armab: Custom Docker image/repository makes sense to move out of "global.yaml" into `st2chatops`, naming for consistency ideally to keep similar to
https://github.com/StackStorm/stackstorm-ha/blob/d8430261c877e1807dcd6f6a96e8dc72b63ffbda/values.yaml#L80-L87

values.yaml
```diff
@@ -355,7 +395,18 @@ st2garbagecollector:
   nodeSelector: {}
   tolerations: []
   affinity: {}
-
+  
+# Chatops service settings  
+st2chatops:
+  replicas: 1
```
armab: Let's remove replicas from configuration as st2chatops can't have more than 1 instances.
Adding a code comment about the fact that st2chatops doesn't scale would be helpful.

chatops-settings.yaml
```diff
@@ -0,0 +1,140 @@
+
```
armab: Similar to https://github.com/StackStorm/st2-dockerfiles/pull/19 `st2chatops.env` is part of the `st2chatops` deb/rpm package and there is no need to include a copy of these vars here.
Especially considering that it could become quickly out of date.

values.yaml
```diff
@@ -355,7 +395,18 @@ st2garbagecollector:
   nodeSelector: {}
   tolerations: []
   affinity: {}
-
+  
+# Chatops service settings  
+st2chatops:
```
armab: I think would be useful to add `enabled: true/false` setting for the whole `st2chatops` as this is something optional, comparing to other required st2 core services.

templates/deployments.yaml
```diff
@@ -106,6 +104,141 @@ spec:
...
+        support: {{ template "supportMethod" . }}
+        chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+        release: {{ .Release.Name }}
+        heritage: {{ .Release.Service }}
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmaps_st2-conf.yaml") . | sha256sum }}
+        checksum/chatops: {{ include (print $.Template.BasePath "/configmaps_chatops.yaml") . | sha256sum }}
+        checksum/packs: {{ include (print $.Template.BasePath "/configmaps_packs.yaml") . | sha256sum }}
+        checksum/auth: {{ include (print $.Template.BasePath "/secrets_st2auth.yaml") . | sha256sum }}
+        checksum/ssh: {{ include (print $.Template.BasePath "/secrets_ssh.yaml") . | sha256sum }}
```
armab: Any ssh configs is not a dependency for `st2chatops` and could be easily removed

templates/deployments.yaml
```diff
@@ -106,6 +104,141 @@ spec:
...
+    spec:
+      {{- if .Values.enterprise.enabled }}
+      imagePullSecrets:
+      - name: {{ .Release.Name }}-st2-license
+      {{- end }}
+      initContainers:
+      {{- if .Values.st2.packs.image.repository }}
+      # Merge packs and virtualenvs from st2actionrunner with those from the st2.packs image
+      # Can be used for chatops pack if system pack comes without it.
```
armab: I'm under impression that access to pack content is not needed for `st2chatops/hubot` itself, as it's something that registered in st2 core.

I think all the section related to custom pack content could be removed from `st2chatops` pod, but can you describe the edge case with the system pack?

templates/deployments.yaml
```diff
@@ -106,6 +104,141 @@ spec:
...
+        - name: st2-rbac-roles-vol
+          configMap:
+            name: {{ .Release.Name }}-st2-rbac-roles
+        - name: st2-rbac-assignments-vol
+          configMap:
+            name: {{ .Release.Name }}-st2-rbac-assignments
+        - name: st2-rbac-mappings-vol
+          configMap:
+            name: {{ .Release.Name }}-st2-rbac-mappings
+        {{- end }}
```
armab: All RBAC-related options can be safely removed, st2chatops component doesn't need it.

values.yaml
mosn: Can you provide an example implementation of this?
the logic is that If a user is left with the responsibility of adding key/values for the custom adapter, then I must find a place to properly document the right values for it.  `values.yaml` is a candidate but then this is will make it crowded and filled with unrelated information.

Another way that I can think of is to separate custom values into its own config file and the user can reference the corresponding configuration file in values.yaml. This way we allow the same configuration file to be used by multiple applications and also we can have documentation for each if the custom values while keeping `values.yaml` nice and clean. 

We can use the same approach for all St2 components. A good implementation of this approach can be seen with   [Gitlan Providers.
](https://docs.gitlab.com/charts/charts/globals.html#providers)

armab: Left an example in https://github.com/StackStorm/stackstorm-ha/pull/55#pullrequestreview-218331459

armab:
Looks like the fork/branch still needs to be synced with upstream properly as Github diff shows unrelated to your PR changes https://github.com/StackStorm/stackstorm-ha/pull/55/files

rapittdev:
Since I'm try

> Looks like the fork/branch still needs to be synced with upstream properly as Github diff shows unrelated to this PR changes https://github.com/StackStorm/stackstorm-ha/pull/55/files

Might it happen because I have local branch set to something other than `master` ? 
I tried...
```
git pull upstream 
```
git remote -v shows 
```
origin	https://github.com/rapittdev/stackstorm-ha.git (fetch)
origin	https://github.com/rapittdev/stackstorm-ha.git (push)
upstream	https://github.com/StackStorm/stackstorm-ha.git (fetch)
upstream	https://github.com/StackStorm/stackstorm-ha.git (push)
```

templates/deployments.yaml
mosn: The only edge case I can think of is when we do not have `chatops` pack installed with other system packs. (not sure if that's even possible) That case can easily render `st2chatops` useless. not to think what will happen to `notifications`

I was originally thinking of a condition where `chatops` pack is not installed and tried to find a way to overcome that scenario.

mosn:
@armab I think we're better off closing this PR when all the details are cleared and submit a new one with proper implementation. If you agree I'll get started with that. This was just an early setup to better get a grasp of what we're dealing with.

rapittdev:
@armab @warrenvw I tried syncing with upstream and after some-time wrestling, I think this is best i got. I am not sure if that's updated since I see merging blocks. But I have an idea. If you agree you can open a new branch for chops so the community can directly contribute to. I think that'll save us time. 

I also was able to deploy to  k8 while pulling the image of the st2-docker file. I added some cleanups on mu code which I will cherry peak into the new branch.

Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 # Update StackStorm version here to rely on other Docker images tags
 appVersion: 3.0dev
 name: stackstorm-ha
-version: 0.10.0
+version: 0.8.4
```
rapittdev: I don't know if the mismatch is because of my mistake or is it normal here

README.md
```diff
@@ -112,27 +112,12 @@ In an HA deployment there must be a minimum of `2` replicas of st2notifier runni
 which in our case is `etcd`.
 
 ### [st2sensorcontainer](https://docs.stackstorm.com/reference/ha.html#st2sensorcontainer)
-st2sensorcontainer manages StackStorm sensors: It starts, stops and restarts them as subprocesses.
-By default, deployment is configured with `1` replica containing all the sensors.
+st2sensorcontainer manages StackStorm sensors: starts, stops and restarts them as a subprocesses.
```
rapittdev: Same I did not add this part.

chatops-settings.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+
+
+
+
+
+export ST2_HOSTNAME="${ST2_HOSTNAME:-localhost}"
```
rapittdev: Looks like I forgot to remove it. Will do that now since we don't need it anymore

templates/configmaps_chatops.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    app: st2chatops
+    tier: backend
+    vendor: stackstorm
+    support: {{ template "supportMethod" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+# see configmaps_st2-conf.yaml for indepth explanantion of mappings here.
```
rapittdev: I figure the implementation is identical to st2.conf without having to apply layers. though I think it is. a good idea to have the first st.conf applies through the system and then users come on topic. As per described previously.

templates/deployments.yaml
```diff
@@ -780,56 +910,56 @@ spec:
...
   # https://docs.stackstorm.com/reference/ha.html#st2sensorcontainer
   # It is possible to run st2sensorcontainer in HA mode by running one process on each compute instance. Each sensor node needs to be
   # provided with proper partition information to share work with other sensor nodes so that the same sensor does not run on different nodes.
   # See Partitioning Sensors for information on how to partition sensors.
+  # TODO: Re-work to use single-sensor-per-container mode instead of running 1 node. Proper implementation is possible with Helm templating (#4)
   replicas: 1
   template:
     metadata:
       labels:
-        app: st2sensorcontainer{{ template "hyphenPrefix" .name }}
```
rapittdev: especially these changes weren't mine. at least something intentional.

armab:
@rapittdev As I see, the github diff still shows old content that was already contributed to upstream `master` branch. We can't open an all-writeable branch for community per Github limitations.
But if you want to open a new PR instead of this one, - that would be fine too.

Chart.yaml
rapittdev: > @rapittdev As I see, the github diff still shows old content that was already contributed to upstream `master` branch. We can't open an all-writeable branch for community per Github limitations.
> But if you want to open a new PR instead of this one, - that would be fine too.

I prefer to preserve the commit history and the flow of conversation. I will give it some time to clean it up tonight.

armab: Sounds good :+1:

mosn:
Sorry for excessive notifications. I tried to sync this fork with the master as much as possible. The current approach is not final but since I received a message from the community members who wanted to contribute, having our fork synced with the master can help them get started easier. 

I was still trying to mount env variable in an easy way for the `st2chatops` docker to use. you're welcome to contribute or use this for inspiration and submit your own PR.

armab:
@mosn Sounds good, I’ll start looking into converging this PR into prod state soon.
But after looking at the changes made, - it’s a great work there! Much more cleaner and stable than before 👍

mosn:
Thanks for the info. I still cannot decide on environment variables  if it’s better to use command arguments or try mounting volumes.  I cannot get the second one to work and I also cannot find a clean way with arguments.

mosn:
@armab if you agree, since this was a playground for chatops integration, I close this PR and you can start from your own branch, I think that will also save time. I continue working on this branch working on Ingress with auto-ssl.

armab:
@mosn Branches are free and it's easy to create one.
If you think you want to close this PR I'm fine with that, but I would prefer to work based on your code, current PR and keeping this discussion history inplace for future context and preserve that contribution credit for you who started `chatops` work.

mosn:
Sounds good to me 👍 let’s keep it here.

rapittdev:
👍👍👍 Thanks @armab  @warrenvw This chart was already ahead, now it's about to get wings!  Thank you for helping us finalize this.  We're planning to launch our product with Stackstorm-Ha!

armab:
@mosn Thank you for starting this K8s Chatops work, making changes and bringing this feature to a working state! Appreciate your contribution :+1: 

Just added a little bit of polishing here and there and tried to find possible corner cases. One of them is reported in https://github.com/StackStorm/st2chatops/issues/124
@warrenvw I'll wait for your additional review in this PR and @mosn if you can double-check :eyes: and verify this PR in action too, - that would help before we merge it early next week.

mosn:
@armab I'll be able to start the integration at 15:00 CET. I will provide updates as I progress through the implementation.

mosn:
Cluster is ready based on following best practice 

https://github.com/bitnami/kube-prod-runtime/blob/master/docs/quickstart-gke.md

Going to start with Stackstorm and chatops now

mosn:
it works great too!

image

@armab I had to move to EWC 3.0 because of this. So far the improved workflow was able to do 5 hours job in 30min. Very helpful with a syntax error. As for chatops I'll finish converting all my mistral chains and then look at that, hopefully, I can work out detailed feedback later in this week.

PR #62, opened 2019-04-09
Generate 'stanley' private SSH key if no one provided in Helm values

### LINKED ISSUE #16: Autogenerate secrets if no Helm value provided, remove insecure defaults

armab:
Closes #16


armab:
After several tries and deeper research, it turns out Helm doesn't provide enough instrumentation to work with autogenerated values and their state.

1) It's possible to autogenerate SSH private key secret, but it will be generated every time on Helm upgrade which is undesired
2) It's possible to workaround SSH private key generation only once on install and mark it as unmanaged by Helm (so it's not deleted on upgrade), but it's not possible to transition to user-provided SSH key in helm values after that
3) It's possible to set SSH private key in values.yaml, but it's not possible to transition to autogenerated if user removed predefined one from Helm values

I've seen other official Helm chart repositories and what they do: allow or autogenerated or predefined secrets without proper transition, meaning user had autogenerated secret on first install (default), they can't change it to custom one having no chance to rotate the secrets afterwards.

With that missing functionality with autogenerated < > predefined value transition I'll keep our Helm values as is until better solution introduced in Helm itself.

There are a lot of discussions around this issue in Helm community and even a WIP PR in links below:
* https://github.com/helm/helm/issues/3053
* https://github.com/helm/charts/issues/5167
* https://github.com/helm/helm/pull/5290

armab:
Per https://github.com/helm/helm/issues/3053#issuecomment-628876519 looks like Helm3 has a solution related to this issue via new `lookup` feature.
We'll need to review this.

cognifloyd:
#203 includes generating the stanley_rsa ssh key by default. So, if that gets merged then this can probably be closed.

Issue #13, opened 2018-09-20
Start mongodb with --auth option


warrenvw:
> Moved from StackStorm/st2enterprise-dockerfiles#62

It is not recommended to run mongodb without authentication, for any length of time.

Some useful references:

 - https://www.mongodb.com/blog/post/running-mongodb-as-a-microservice-with-docker-and-kubernetes
 - http://pauldone.blogspot.com/2017/06/deploying-mongodb-on-kubernetes-gke25.html
 - https://www.serverlab.ca/tutorials/linux/database-servers/how-to-create-mongodb-replication-clusters/

Chart has `auth` section to configure: https://github.com/helm/charts/tree/master/stable/mongodb-replicaset#authentication so should be doable with Helm `values.yaml`.

armab:
Implemented via https://github.com/StackStorm/stackstorm-ha/pull/63

PR #63, opened 2019-04-12
Add mongo auth

Lomonosow:
#13  Is work for me, but use mongo admin user. I`d tried to find another way but unsuccessfully)

### DIFF
--- a/templates/configmaps_st2-conf.yaml
+++ b/templates/configmaps_st2-conf.yaml
@@ -27,7 +27,11 @@ data:
     url = amqp://{{ required "rabbitmq-ha.rabbitmqUsername is required!" (index .Values "rabbitmq-ha" "rabbitmqUsername") }}:{{ required "rabbitmq-ha.rabbitmqPassword is required!" (index .Values "rabbitmq-ha" "rabbitmqPassword") }}@{{ .Release.Name }}-rabbitmq-ha-discovery:5672
     [database]
     # TODO: ReplicaSet connection string needs templating based on number of MongoDB nodes
+    {{- if index .Values "mongodb-ha" "auth" "enabled" }}
+    host = mongodb://{{ required "mongodb-ha.auth.adminUser is required!" (index .Values "mongodb-ha" "auth" "adminUser") }}:{{ required "mongodb-ha.auth.adminPassword is required!" (index .Values "mongodb-ha" "auth" "adminPassword") }}@{{ .Release.Name }}-mongodb-ha/?authSource=admin&replicaSet=rs0
+    {{ else }}
     host = mongodb://{{ .Release.Name }}-mongodb-ha/?replicaSet=rs0
+    {{- end }}
 
   # User-defined st2 config with custom settings applied on top of everything else.
   # The order of merging: st2.conf < st2.docker.conf < st2.user.conf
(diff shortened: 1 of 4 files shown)


warrenvw:
@Lomonosow Thank you for your contribution. I'll be reviewing it shortly.

Lomonosow:
I found some mistake in configuration, mongo want key at least 6 symbol length. I have fix it in my commit.

armab:
@warrenvw Can you please drive this review, think about corner cases and check the PR?

From my point of view it would be nice to add a condition and set mongo username/password in `st2.conf` only if `mongodb-ha.auth.enabled = true`

templates/configmaps_st2-conf.yaml
```diff
@@ -27,7 +27,7 @@ data:
     url = amqp://{{ required "rabbitmq-ha.rabbitmqUsername is required!" (index .Values "rabbitmq-ha" "rabbitmqUsername") }}:{{ required "rabbitmq-ha.rabbitmqPassword is required!" (index .Values "rabbitmq-ha" "rabbitmqPassword") }}@{{ .Release.Name }}-rabbitmq-ha-discovery:5672
     [database]
     # TODO: ReplicaSet connection string needs templating based on number of MongoDB nodes
-    host = mongodb://{{ .Release.Name }}-mongodb-ha/?replicaSet=rs0
+    host = mongodb://{{ required "mongodb-ha.auth.adminUser is required!" (index .Values "mongodb-ha" "auth" "adminUser") }}:{{ required "mongodb-ha.auth.adminPassword is required!" (index .Values "mongodb-ha" "auth" "adminPassword") }}@{{ .Release.Name }}-mongodb-ha/?authSource=admin&replicaSet=rs0
```
warrenvw: Per @armab, add a condition and set mongo username/password in `st2.conf` only if `mongodb-ha.auth.enabled = true`.

values.yaml
```diff
@@ -367,6 +367,11 @@ mongodb-ha:
   image:
     # StackStorm currently supports maximum MongoDB v3.4
     tag: 3.4
+  auth:
+    enabled: true
+    adminUser: "admin"
+    adminPassword: "admin"
+    key: "keykeykey"
```
warrenvw: I would suggest adding a note that the default `key` should be changed, similar to the note above `secrets`.

`NB! It's highly recommended to change ALL defaults!`

Lomonosow: Added

templates/configmaps_st2-conf.yaml
Lomonosow: Done

warrenvw:
I'd appreciate if you could update the Changelog and bump the version in `Charts.yaml`. The Changelog should probably indicate that auth is now enabled by default.

Please note that I have not been able to successfully transition mongodb from auth disabled to auth enabled. Have you tried this? I'm attempting this particular scenario because auth is likely disabled on most installations... and now by default it will be enabled.

For example, in the `st2actionrunner` pod, I see:

```
OperationFailure: Authentication failed.
```

Lomonosow:
> For example, in the st2actionrunner pod, I see:
> 
> OperationFailure: Authentication failed.
I have reproduce it when I`d change mongodb admin password and tried to upgrade helm release. Mongodb chart does not recreate user if directory for db already exists (in mongodb persistent storage enabled by default and if you delete release by helm, persistent storage will not be deleted).

PR #64, opened 2019-04-13
Add missing documentation for st2scheduler

armab:
Copy missing `scheduler` documentation from st2docs https://docs.stackstorm.com/install/k8s_ha.html#st2scheduler to `stackstorm-ha`

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,8 @@ All the workflow engine processes will share the load and pick up more work if o
 >  [Orquesta st2workflowengine](https://docs.stackstorm.com/orchestra/index.html) as a new native workflow engine.
 
 ### [st2scheduler](https://docs.stackstorm.com/reference/ha.html#st2scheduler)
-TODO: Description TBD
+`st2scheduler` is responsible for handling ingress action execution requests.
+`2` replicas for K8s Deployment are configured by default to increase StackStorm scheduling throughput.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,7 +103,8 @@ All the workflow engine processes will share the load and pick up more work if o
 >  [Orquesta st2workflowengine](https://docs.stackstorm.com/orchestra/index.html) as a new native workflow engine.
 
 ### [st2scheduler](https://docs.stackstorm.com/reference/ha.html#st2scheduler)
-TODO: Description TBD
+`st2scheduler` is responsible for handling ingress action execution requests.
+`2` replicas for K8s Deployment are configured by default to increase StackStorm scheduling throughput.
```
warrenvw: Maybe start with a word instead? I suggest: "By default, 2 replicas are configured in the K8s deployment to increase StackStorm scheduling throughput."

armab: Seems very minor.

The issue is that it's extracted from the https://docs.stackstorm.com/install/k8s_ha.html so I'm trying to keep both docs in sync.
Any wording change would need to be done in both places.

warrenvw: Yes, it's definitely minor. I am okay leaving it as is if you'd prefer.

armab: Thanks, merging then!

PR #66, opened 2019-04-16
Move st2web.annotations into st2web.service

warrenvw:
Move `st2web.annotations` to `st2web.service.annotations`, to match `values.yaml`.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 # Update StackStorm version here to rely on other Docker images tags
 appVersion: 3.0dev
 name: stackstorm-ha
-version: 0.11.0
+version: 0.12.0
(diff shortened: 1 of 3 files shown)


armab:
@warrenvw Is there anything that blocks to merge this and release a new patched chart version?

Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 # Update StackStorm version here to rely on other Docker images tags
 appVersion: 3.0dev
 name: stackstorm-ha
-version: 0.11.0
+version: 0.12.0
```
armab: Just a not important note for future: ideal here would be to bump patch version aka `0.11.1` since it's a small fix, no new functionality introduced.

warrenvw: Considered this, but incorrectly assumed change was user visible. It is not. So `0.11.1` was the better choice. Next time I will look at the change more closely.

PR #67, opened 2019-05-10
Use st2 3.1dev

warrenvw:
Since st2 `3.0.0` has been released, bump appVersion to `3.1dev`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## In Development
 
-## v0.13.1
+## v0.14.0
 * Pin st2 version to `v3.1dev` as a new latest development version (#67)
 
 ## v0.13.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## In Development
 
+## v0.13.1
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## In Development
 
+## v0.13.1
+* Pin st2 version to `v3.1dev` as a new latest development version (#67)
```
armab: I think `0.14.0` would be more appropriate versioning since it's not a bugfix or patch but a version change.

https://github.com/StackStorm/stackstorm-ha/blob/master/CHANGELOG.md#v083 and https://github.com/StackStorm/stackstorm-ha/blob/master/CHANGELOG.md#v083 were a bit messy due to updated release plans changes and version renames.

warrenvw: I was going to do a minor version bump for the reasons you mention, but I noticed that `v0.8.4` just bumped patch when changing appVersion from `v2.10dev` to `3.0dev`. Figured it set a precedent...

Anyhow, I agree that we shouldn't bump patch when appVersion is changed. I'll make the change you suggest.

PR #68, opened 2019-05-17
Add support for ingress

warrenvw:
See #6 for more details. The following change is made in this PR:

> Expose Ingress controller settings via Helm `values.yaml` to allow users to configure the SSL/TLS negotiation layer on their own (optional).

The ALB is configured using annotations. For example:

```
annotations:
  kubernetes.io/ingress.class: alb
  alb.ingress.kubernetes.io/backend-protocol: HTTP
  alb.ingress.kubernetes.io/scheme: internet-facing
  alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS": 443}]'
  alb.ingress.kubernetes.io/certificate-arn: ******
  ...
```

Remaining tasks:
- [ ] Additional testing
- [x] Update `Chart.yaml` with next minor `version`.
- [x] Update `CHANGELOG.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,17 @@ See Helm `values.yaml`, `enterprise` section for configuration examples.
 
 The default configuration values for this chart are described in `values.yaml`.
 
+## Ingress
+
+Use an ingress if you need to expose HTTP and HTTPS routes from outside the cluster to services
+within the cluster. Ingress provides load balancing, SSL termination and name-based virtual
+hosting. See the ingress section in `values.yaml` for configuration details.
+
+You will need to deploy an ingress controller such as
+https://github.com/helm/charts/tree/master/incubator/aws-alb-ingress-controller or
+https://www.nginx.com/products/nginx/kubernetes-ingress-controller. More are available
+at https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/.
+
 ## Components
 
 The Community FOSS Dockerfiles used to generate the docker images for each st2 component are available at
--- a/README.md
+++ b/README.md
@@ -44,6 +44,10 @@ See Helm `values.yaml`, `enterprise` section for configuration examples.
 > Don't have StackStorm Enterprise License?<br>
 > 90-day free trial can be requested at https://stackstorm.com/#product
 
+## Configuration
+
+The default configuration values for this chart are described in `values.yaml`.
+
 ## Components
 
 The Community FOSS Dockerfiles used to generate the docker images for each st2 component are available at
--- a/README.md
+++ b/README.md
@@ -50,14 +50,15 @@ The default configuration values for this chart are described in `values.yaml`.
 
 ## Ingress
 
-Use an ingress if you need to expose HTTP and HTTPS routes from outside the cluster to services
-within the cluster. Ingress provides load balancing, SSL termination and name-based virtual
-hosting. See the ingress section in `values.yaml` for configuration details.
+Ingress is worth considering if you want to expose multiple services under the same IP address, and
+these services all use the same L7 protocol (typically HTTP). You only pay for one load balancer if
+you are using native cloud integration, and because Ingress is "smart", you can get a lot of
+features out of the box (like SSL, Auth, Routing, etc.). See the ingress section in `values.yaml`
+for configuration details.
 
-You will need to deploy an ingress controller such as
-https://github.com/helm/charts/tree/master/incubator/aws-alb-ingress-controller or
-https://www.nginx.com/products/nginx/kubernetes-ingress-controller. More are available
-at https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/.
+You will first need to deploy an ingress controller of your preference. See
+https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/#additional-controllers
+for more information.
 
 ## Components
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -44,6 +44,21 @@ See Helm `values.yaml`, `enterprise` section for configuration examples.
...
+## Ingress
+
+Use an ingress if you need to expose HTTP and HTTPS routes from outside the cluster to services
+within the cluster. Ingress provides load balancing, SSL termination and name-based virtual
+hosting. See the ingress section in `values.yaml` for configuration details.
+
+You will need to deploy an ingress controller such as
+https://github.com/helm/charts/tree/master/incubator/aws-alb-ingress-controller or
+https://www.nginx.com/products/nginx/kubernetes-ingress-controller. More are available
+at https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/.
```
armab: I think listing so many URLs is excessive. 

Saying that user will need to deploy & configure an Ingress controller of his preference beforehand would be enough. I'm good with including https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/#additional-controllers if you think that can help.

README.md
```diff
@@ -44,6 +44,21 @@ See Helm `values.yaml`, `enterprise` section for configuration examples.
 > Don't have StackStorm Enterprise License?<br>
 > 90-day free trial can be requested at https://stackstorm.com/#product
 
+## Configuration
+
+The default configuration values for this chart are described in `values.yaml`.
+
+## Ingress
+
+Use an ingress if you need to expose HTTP and HTTPS routes from outside the cluster to services
```
armab: Exposing could be done with `LoadBalancer` or with `NodePort` as well. Explanation behind the need of Ingress needs some re-wording.

templates/ingress.yaml
```diff
@@ -1 +1,37 @@
-# TODO: Research & add Ingress controller for exposing st2web service to public net (#6)
+{{- if .Values.ingress.enabled }}
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  {{- $name := printf "%s-%s" $.Release.Name .Values.ingress.name | trunc 63 | trimSuffix "-" | quote }}
```
armab: We don't have `Values.ingress.name` defined anywhere

templates/ingress.yaml
```diff
@@ -1 +1,37 @@
...
+  name: {{ $name }}
+  labels:
+    app: {{ $name }}
+    chart: "{{ $.Chart.Name }}-{{ $.Chart.Version }}"
+    release: "{{ $.Release.Name }}"
+    heritage: "{{ $.Release.Service }}"
```
armab: Please follow the `labels` from the other K8s resources.
There are `tier`, `vendor` and `support` labels missing

templates/ingress.yaml
```diff
@@ -1 +1,37 @@
...
+  rules:
+  {{- range .Values.ingress.hosts }}
+  - host: {{ .name }}
+    http:
+      paths:
+      {{- range .paths }}
+        - path: {{ default "/*" .path }}
+          backend:
+            serviceName: {{ .serviceName }}
+            servicePort: {{ .servicePort }}
```
armab: serviceName/servicePort should be optional and default to templated version of st2web service we already deploy in the cluster.
This way user won't need to guess anything from the `values.yaml` perspective or even worse hardcode releaseName from values.yaml because we didn't template it properly here.

templates/ingress.yaml
armab: Having no `Values.ingress.name`, we can switch back to simplified templating similar to
https://github.com/StackStorm/stackstorm-ha/blob/2f92c2679948ca9bd7468e015571bfb1b96b5149/templates/services.yaml#L5

README.md
```diff
@@ -44,6 +44,21 @@ See Helm `values.yaml`, `enterprise` section for configuration examples.
 > Don't have StackStorm Enterprise License?<br>
 > 90-day free trial can be requested at https://stackstorm.com/#product
 
+## Configuration
+
+The default configuration values for this chart are described in `values.yaml`.
+
+## Ingress
```
armab: Just a reminder that after merging this PR we'll need to sync the respective `st2docs` for K8s/HA

templates/ingress.yaml
```diff
@@ -1 +1,49 @@
...
+      {{- end }}
+  {{- else }}
+    {{- if required "Missing context '.Values.st2web.service.hostname'!" .Values.st2web.service.hostname }}
+  - hosts: {{ .Values.st2web.service.hostname }}
```
armab:
```suggestion
  - host: {{ .Values.st2web.service.hostname }}
```

values.yaml
```diff
@@ -120,6 +120,30 @@ st2:
...
+  annotations: {}
+    # kubernetes.io/ingress.class: nginx
+    # kubernetes.io/tls-acme: "true"
+  # Map hosts to paths
+  hosts: []
+  # - host: hostname.domain.tld
+  #   # Map paths to services
+  #   paths:
+  #     - path: /
+  #       serviceName: service
```
armab: Worth adding that both path and service details are optional and that working defaults will be used to automatically configure st2web for ingress

warrenvw:
Corner cases tested. All looks good.

armab:
:rocket:

PR #70, opened 2019-05-24
Setup K8s infrastructure for Helm e2e tests

armab:
Part of the #27 

This PR runs Helm install in CI as a basic check to verify if chart can be deployed successfully on a real cluster.

Uses `KinD`, - Kubernetes in Docker which is lightweight way to setup K8s cluster for a quick CI/testing purposes. See:
- https://github.com/kubernetes-sigs/kind
- https://kind.sigs.k8s.io/

Test which one is more KinD-stable?
- [x] CircleCI (https://circleci.com/gh/StackStorm/stackstorm-ha/458)
- [x] TravisCI (https://travis-ci.org/StackStorm/stackstorm-ha/jobs/537943346)


armab:
Alright, Helm chart e2e CI test blueprint works in both CircleCI and TravisCI with K8s deployed by KinD.

Deployment within KinD is not happy about persistent volumes coming from the official MongoDB-HA Helm chart and so I had to disable persistence.
I also tried TravisCI to verify if it's something KinD environment specific or CI platform related. Travis has similar problems or workarounds comparing to CircleCI so we'll stick with Circle as initially chosen CI for other tasks.

Overall it's great that we can run quick Helm e2e tests in something like CircleCI instead of spinning real K8s cluster in cloud.

armab:
Good for now, I'll continue adding Helm e2e tests in another PR

.circleci/config.yml
```diff
@@ -1,4 +1,12 @@
-version: 2
+version: 2.1
+
+# Add additional CircleCI Orbs dependencies
+orbs:
```
warrenvw: Hoorah! I do believe this is the first use of orbs! 👍

armab: Was good to avoid some wget binary + chmod scripting in favor of these.

PR #72, opened 2019-05-29
Replace st2web https with http

warrenvw:
See https://github.com/StackStorm/st2-dockerfiles/pull/20 for more context.

Since `st2web` backend now defaults to using `HTTP` instead of `HTTPS`, remove all the `HTTPS` related stuff, including the certs. Update ports.

Remaining tasks:

 - [x] Review code to see if any additional changes need to be made.
 - [x] Ensure st2web can be reached when `ST2WEB_HTTPS=0`.


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## In Development
 
+## v0.16.0
+* st2web now uses HTTP by default (#72)
```
armab: Ideally to add a few more words that it's now recommended to rely on "LoadBalancer" or "Ingress" to add HTTPs layer on top of it.

warrenvw: That's a good idea. Thanks.

README.md
```diff
@@ -83,6 +83,8 @@ kubectl exec -it ${ST2CLIENT} /bin/bash
 ### [st2web](https://docs.stackstorm.com/latest/reference/ha.html#nginx-and-load-balancing)
 st2web is a StackStorm Web UI admin dashboard. By default, st2web K8s config includes a Pod Deployment and a Service.
 `2` replicas (configurable) of st2web serve the web app and proxy requests to st2auth, st2api, st2stream.
+By default, st2web uses HTTP instead of HTTPS. Pass in ST2WEB_HTTPS environment variable with a non-zero value
+(e.g. "1") and ensure SSL certificates are configured.
```
armab: I don't think we need to mention `ST2WEB_HTTPS` or certificates here as it was designed for possible docker-compose use cases outside of K8s where users don't have concept of Ingress layer.

README.md
```diff
@@ -83,6 +83,8 @@ kubectl exec -it ${ST2CLIENT} /bin/bash
...
 > **Note!** By default, st2web is a NodePort Service and is not exposed to the public net.
   If your Kubernetes cluster setup supports the LoadBalancer service type, you can edit the corresponding helm values to configure st2web as a LoadBalancer service in order to expose it and the services it proxies to the public net.
```
armab: Can mention Ingress here as well.

Issue #28, opened 2018-10-09
Helm chart Unit Tests


armab:
Apart of Helm Chart Integration Testing in #27 which could be ran only against real K8s cluster and deployed release, consider adding Unit Tests for smaller item checks which is available as a Helm plugin: https://github.com/lrills/helm-unittest

cognifloyd:
Though not a unittest, I just experimented with using https://github.com/bats-core/bats-detik to have the inegration tests query k8s to ensure things are setup correctly. Seems to work ok.

cognifloyd:
I looked into the helm-unittest plugin. There are a chain of repos where one person and then another is maintaining it. And I don't relish writing unit tests in YAML even though everything else is in YAML.

Looking farther, it looks like the airflow folks pioneered using pytest to unit test helm charts:
https://github.com/helm-unittest/helm-unittest/issues/110

Luckily airflow (and in particular the helm chart + tests) is [Apache 2.0 licensed](https://github.com/apache/airflow/blob/main/chart/LICENSE). So, we could use this utility file and start writing pytest-style unit tests.
https://github.com/apache/airflow/blob/main/chart/tests/helm_template_generator.py

cognifloyd:
I've added a bunch of tests in #288.

Other tests that I think would be good (I'm listing values) in no particular order:
- env and envFromSecrets
    - `st2*.env` and `jobs.env`
    - `st2actionrunner.envFromSecrets`,
    - `st2sensorcontainer.envFromSecrets`,
    - `st2client.envFromSecrets`,
    - and `jobs.envFromSecrets`
- `st2*.extra_volumes` (all except st2chatops have it)
- `st2.packs.volumes`
- st2chatops
    - `st2chatops.enabled`
    - `st2chatops.env` (some special handling of api key var)
    - `st2chatops.hubotScriptsVolume`
- config
    - `st2.config`
    - `st2client.st2clientConfig`
    - `jobs.st2clientConfig`
- HTTPS in st2web (`st2web.env.ST2WEB_HTTPS` and `st2web.extra_volumes`)
- image
    - `image.repository`
    - `image.tag`
    - `st2*.image.tag` and `jobs.image.tag`
- `st2*.replicas` (some things can only ever have 1, others can have multiple...)
- st2 security
    - `st2.username` and `st2.password`
    - `st2.system_user.user` and `st2.system_user.ssh_key_file`
- `st2.datastore_crypto_key`
- `st2.rbac`
[...]

PR #80, opened 2019-08-26
Compose MongoDB-HA cluster connection string for many nodes

armab:
Rely on multi-node connection string for MongoDB-HA cluster instead of using loadbalancer single host which is not desired for prod deployment.

See https://api.mongodb.com/python/current/examples/high_availability.html#mongos-load-balancing
> Warning
> Do not connect PyMongo to a pool of mongos instances through a load balancer. A single socket connection must always be routed to the same mongos instance for proper cursor support.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -32,3 +32,28 @@ stackstorm
 {{- define "hyphenPrefix" -}}
 {{ if . }}-{{ . }}{{end}}
 {{- end -}}
+
+# Allow calling helpers from nested sub-chart
+# https://stackoverflow.com/a/52024583/4533625
+# https://github.com/helm/helm/issues/4535#issuecomment-477778391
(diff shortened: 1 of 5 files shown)


templates/_helpers.tpl
```diff
@@ -32,3 +32,28 @@ stackstorm
...
+# Usage: "{{ include "nested" (list . "mongodb-ha" "mongodb-replicaset.fullname") }}"
+{{- define "nested" }}
+{{- $dot := index . 0 }}
+{{- $subchart := index . 1 | splitList "." }}
+{{- $template := index . 2 }}
+{{- $values := $dot.Values }}
+{{- range $subchart }}
+{{- $values = index $values . }}
+{{- end }}
+{{- include $template (dict "Chart" (dict "Name" (last $subchart)) "Values" $values "Release" $dot.Release "Capabilities" $dot.Capabilities) }}
```
armab: All this trickery should be eventually covered by respective Unit Tests https://github.com/StackStorm/stackstorm-ha/issues/28

PR #81, opened 2019-08-26
Use stable etcd operator as opposed to etcd from incubator

### LINKED ISSUE #75: etcd deployment is unstable
We're using external chart dependency https://github.com/helm/charts/tree/master/incubator/etcd from Helm incubator to deploy `etcd` cluster as part of the `stackstorm-ha`.

It doesn't work well, failing frequently, like https://circleci.com/gh/StackStorm/stackstorm-ha/604.

Example K8s pod list:
```
NAME                                                              READY   STATUS             RESTARTS   AGE
st2cicd-etcd-0                                                    0/1     CrashLoopBackOff   107        9h
st2cicd-etcd-1                                                    1/1     Running            0          11h
st2cicd-etcd-2                                                    1/1     Running            112        9h
```

With error log from `etcd` pod:
```
Waiting for st2cicd-etcd-2.st2cicd-etcd to come up
Re-joining etcd member
cat: can't open '/var/run/etcd/member_id': No such file or directory
```

Research other solutions, consider `etcd` operator and how to automate that via Helm deployment.

trstruth:
Closes #75 - How can we test that st2 is still stable with a different dependency?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## v0.20.0
 * Add option to disable MongoDB and RabbitMQ in-cluster deployment and configuration (#79)
 * Compose multi-node connection string for MongoDB cluster instead of using loadbalancer single host (#80)
+* Change etcd dependency from incubator/etcd to stable/etcd-operator (#81)
 
 ## v0.19.0
 * Configure RabbitMQ Queue mirroring by default, see https://www.rabbitmq.com/ha.html (#78)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,12 @@
 
 ## In Development
 
+## v0.21.0
+* Change etcd dependency from incubator/etcd to stable/etcd-operator (#81)
+
 ## v0.20.0
 * Add option to disable MongoDB and RabbitMQ in-cluster deployment and configuration (#79)
 * Compose multi-node connection string for MongoDB cluster instead of using loadbalancer single host (#80)
-* Change etcd dependency from incubator/etcd to stable/etcd-operator (#81)
 
 ## v0.19.0
 * Configure RabbitMQ Queue mirroring by default, see https://www.rabbitmq.com/ha.html (#78)
--- a/templates/configmaps_st2-conf.yaml
+++ b/templates/configmaps_st2-conf.yaml
@@ -21,7 +21,7 @@ data:
     [auth]
     api_url = http://{{ .Release.Name }}-st2api{{ template "enterpriseSuffix" . }}:9101/
     [coordination]
-    url = etcd://{{ .Release.Name }}-etcd:2379
+    url = etcd://{{ index .Values "etcd-operator" "etcdCluster" "name" }}-client:2379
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## v0.20.0
 * Add option to disable MongoDB and RabbitMQ in-cluster deployment and configuration (#79)
 * Compose multi-node connection string for MongoDB cluster instead of using loadbalancer single host (#80)
+* Change etcd dependency from incubator/etcd to stable/etcd-operator (#81)
```
armab: `v0.20.0` was already released yesterday https://github.com/StackStorm/stackstorm-ha/releases/tag/v0.20.0

Let's tag it `v0.21.0` and also bump version in `Chart.yaml`

armab:
CI is also unhappy. 

If you look at CI (https://circleci.com/gh/StackStorm/stackstorm-ha/907) it shows that some st2 services pods are in Crashloop. Most probably they can't connect to new etcd backend due to changed endpoint hostname or similar.

armab:
After looking at it closer, `etcd-operator` stable Helm chart by default doesn't create the etcd cluster itself.

Only by providing Helm value `customResources.createEtcdClusterCRD = true` it can create a cluster, but with that K8s CustomResourceDefinition named `EtcdCluster` is not available at first run:

```
helm install .
Error: validation failed: unable to recognize "": no matches for kind "EtcdCluster" in version "etcd.database.coreos.com/v1beta2"
```

> https://github.com/helm/charts/tree/master/stable/etcd-operator#installing-the-chart
> Note: If you set cluster.enabled on install, it will have no effect. Before you create an etcd cluster, the TPR must be installed by the operator, so this option is ignored during helm installs, but can be used in upgrades.


So the problem with this Helm chart is that it can't be used automatically as dependency since needed CustomResourceDefinition is not available during the install time, meaning etcd-operator Helm chart needs to be applied manually before `stackstorm-ha`. Not an option.

armab:
An alternative is to try to improve `etcd-operator` Helm chart https://github.com/helm/charts/tree/master/stable/etcd-operator by defining CustomResourceDefinition in Helm template. 
To make sure CRD is registered before we can use it, Helm provides `custom-resource-defintion` pre-hook as described in: https://helm.sh/docs/developing_charts/#defining-a-crd-with-the-crd-install-hook

A quick example from `prometheus-operator`:
https://github.com/helm/charts/blob/32cfeb0527789f2354d4bc7ce978a2e67992e014/stable/prometheus-operator/templates/prometheus-operator/crd-prometheusrules.yaml#L15

templates/configmaps_st2-conf.yaml
```diff
@@ -21,7 +21,7 @@ data:
     [auth]
     api_url = http://{{ .Release.Name }}-st2api{{ template "enterpriseSuffix" . }}:9101/
     [coordination]
-    url = etcd://{{ .Release.Name }}-etcd:2379
+    url = etcd://{{ index .Values "etcd-operator" "etcdCluster" "name" }}-client:2379
```
armab: Updated coordination backend connection string for new etcd cluster

PR #82, opened 2019-08-27
fix bug in template line 46 = should be :=

cmmdrdata:


templates/_helpers.tpl
```diff
@@ -43,7 +43,7 @@ stackstorm
 {{- $template := index . 2 }}
 {{- $values := $dot.Values }}
 {{- range $subchart }}
-{{- $values = index $values . }}
+{{- $values := index $values . }}
```
armab: I believe existing code behaves as intended.
The variable was declared before, but gets re-assigned via `=` in the loop.

Also see https://golang.org/doc/go1.11#text/template

cmmdrdata: it failed for me when I ran it with the error pointing to this line..  after i changed it, it worked.

armab: What was the Helm version you used?
Is there an exact error log you can share?

cmmdrdata: Error: parse error in \"stackstorm-ha/templates/_helpers.tpl\": template: stackstorm-ha/templates/_helpers.tpl:46: unexpected \"=\" in operand"

armab: What's your `helm version` ?

cmmdrdata: Client: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}

armab: Helm `v2.11.0` was released 1 year ago with K8s `v1.11` that reached EOL.

As I checked the closest Helm version that doesn't have templating issue you're experiencing is `v2.12.0`. 
If you can just upgrade to that next version, - that would fix it. Better yet - to latest Helm, if possible.

cmmdrdata:
Ok,  wil do
Sent from Yahoo Mail on Android 
 
  On Fri, Aug 30, 2019 at 9:22, Eugen C.<notifications@github.com> wrote:   
@armab commented on this pull request.

In templates/_helpers.tpl:
> @@ -43,7 +43,7 @@ stackstorm
 {{- $template := index . 2 }}
 {{- $values := $dot.Values }}
 {{- range $subchart }}
-{{- $values = index $values . }}
+{{- $values := index $values . }}

Helm v2.11.0 was released 1 year ago with K8s v1.11 that reached EOL.

As I checked the closest Helm version that doesn't have templating issue you're experiencing is v2.12.0.
If you can just upgrade to that next version, - that would fix it. Better yet - to latest Helm, if possible.

armab:
Closing as workaround provided. We also can't support really old Helm versions in stackstorm-ha beta of the chart.

Issue #89, opened 2019-10-19
Helm v2.15.0 stopped templating MongoDB connection string


armab:
With new released Helm `v2.15.0` templating used to populate MongoDB host stopped working:

```diff
[database]
- host = mongodb://test-mongodb-ha-0.test-mongodb-ha,test-mongodb-ha-1.test-mongodb-ha,test-mongodb-ha-2.test-mongodb-ha/?authSource=admin&replicaSet=rs0
+ host = mongodb:///?authSource=admin&replicaSet=rs0

```
resulting in failed st2 Pods and red e2e builds https://circleci.com/gh/StackStorm/stackstorm-ha/1147

Something might be changed with the Helm template engine. Needs investigation and replacing template rule to populate the MongoDB host again.

armab:
Fixed in upstream https://github.com/StackStorm/stackstorm-ha/pull/91 with Helm `v2.15.1`.

PR #91, opened 2019-10-19
Fix MongoDB connection string templating for Helm 2.15

armab:
Fixes https://github.com/StackStorm/stackstorm-ha/issues/89 where Helm templating failed to generate connection string for MongoDB cluster.

For some reason `index .Values "mongodb-ha" "replicas"` now returns `json.Number` instead of `float64` and so conversion failed in Helm `v2.15`, previously worked in `v2.14` (See https://github.com/helm/helm/issues/6708 for details).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## In Development
 
+## v0.22.1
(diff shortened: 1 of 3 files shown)


armab:
Fixed in upstream https://github.com/StackStorm/stackstorm-ha/pull/91 with Helm `v2.15.1`.

PR #107, opened 2020-01-28
Fix failed CI builds due to e2e timeouts/resources issue

armab:
StackStorm-HA deployment timeouts due to high amount of resources required to provision MongoDB/RabbitMQ/etcd cluster as well as `25+` st2 pods. It all exceeds allowed 10mins timeout in CircleCI builds.

Try to use `Large` CircleCI machine type with more vCPUs & RAM: https://circleci.com/docs/2.0/configuration-reference/#machine-executor-linux

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,6 +66,8 @@ jobs:
 
   # Spin up minikube K8s cluster and run Helm chart & e2e tests on it
   helm-e2e:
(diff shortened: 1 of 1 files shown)


armab:
Helm Deployment task which didn't complete in `10 mins` previously, now completes in `4 mins`.
This definitely works better.

armab:
Hello to the green :heavy_check_mark:  builds again.

PR #108, opened 2020-02-10
Added templating capability for st2kv from values

### LINKED ISSUE #102: Need to use templating on st2kv
As you know, you cannot use templating on ```values.yaml```. I would like to put stackstorm-ha in the chart that I have written in ```requirements.yaml``` but I have a problem that I cannot pass on. I must hold some URL information such as DB or any kind of service URL but while doing that, I have to use {{ .Release.name }} while doing that and I cannot put release name  in values.yaml statically. 

I have figured a way but I can only do it by modifying ```secrets_st2kv.yaml``` which is to use a placeholder for the release name and then replace that placeholder with the release name on ```secrets_st2kv.yaml``` like this:

```
data:
  {{- $var := .Release.Name }}
  st2kv.yaml: {{ toYaml .Values.st2.keyvalue | replace "CHANGEFORRELEASE" $var | b64enc | quote }}
```

This works just fine, but if I do this, I will lose the ability to use stackstorm-ha as a dependency for my chart. I would like to hear your opinions about this. I really need this in someway and I would like not to break a good practice.

erenatas:
Closes #102

With this commit, a user should be able to use templating features in `st2.keyvalue`.

Example usage:
<pre>
  keyvalue:
    - name: xxx_mongo_ipv4
      scope: st2kv.system
      secret: false
      encrypted: false
      value: "<b>{{ .Release.Name }}</b>-xxx-mongodb-ha-client"
</pre>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## In Development
-* Add ability of templating on `st2.kv`
+* Add ability of templating on `st2.keyvalue` in Helm Values (#108)
 
 ## v0.23.0
 * Add support for latest K8s version `1.16`, update e2e CI
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## In Development
+* Add ability of templating on `st2.kv`
 
 ## v0.23.0
 * Add support for latest K8s version `1.16`, update e2e CI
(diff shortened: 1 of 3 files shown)


erenatas:
I hope it's fine now. I can also write the templating ability as a comment instead of a st2kv example to reduce the lines if you would like.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## In Development
+* Add ability of templating on `st2.kv`
```
```suggestion
* Add ability of templating on `st2.keyvalue` in Helm Values (#108)
```

erenatas:
> LGTM. Thanks! 
> 
> Just left one minor comment to address

Done! Thank you so much :)

Issue #110, opened 2020-02-12
Ingress servicePort is defined as String


erenatas:
I believe I have missed this one. 

https://github.com/StackStorm/stackstorm-ha/blob/ff078b22af1e31b175957938d11405b3163778d1/templates/ingress.yaml#L43

Should not be defined as a string, which results in the following error:

```
helm install test .
manifest_sorter.go:175: info: skipping unknown hook: "crd-install"
Error: Ingress.extensions "test-ingress" is invalid: spec.rules[0].http.backend.servicePort: Invalid value: "80": must contain at least one letter or number (a-z, 0-9)
```

armab:
Is this with Helm 3.0?

erenatas:
> Do you get this error with Helm 3.0?

Yes.

armab:
OK, stackstorm-ha doesn't officially support Helm 3 yet, but future fixes like that would be nice to have!

PR #111, opened 2020-02-12
Fix/Ingress servicePort

erenatas:
This PR fixes the ingress service port, as it's issue has been defined here:

Fixes https://github.com/StackStorm/stackstorm-ha/issues/110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## In Development
-* Fix st2web ingress which should have been defined as an Integer instead of a String.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## In Development
+* Fix st2web ingress which should have been defined as an Integer instead of a String.
```
```suggestion
* Fix st2web ingress which should have been defined as an Integer instead of a String (#111)
```

Now that we have a PR #, would be nice to reference it in the Changelog

PR #112, opened 2020-02-14
Update ingress.yaml

erenatas:
Using `{{ .Release.Name }}-ingress` is not informing enough and causes a problem when using `stackstorm-ha` as a dependency for another chart. Helm creates the ingress resource as `{{ .Release.Name }}-ingress` but it's not the ingress of the current chart. This would make the ingress resource of `stackstorm-ha` more definitive.

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -3,7 +3,7 @@
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
-  name: {{ .Release.Name }}-ingress{{ template "enterpriseSuffix" . }}
+  name: {{ .Release.Name }}-st2web-ingress{{ template "enterpriseSuffix" . }}
   labels:
     app: ingress
     tier: frontend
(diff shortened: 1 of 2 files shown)


templates/ingress.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
-  name: {{ .Release.Name }}-ingress{{ template "enterpriseSuffix" . }}
+  name: {{ .Release.Name }}-st2web-ingress{{ template "enterpriseSuffix" . }}
```
armab: To follow the same naming patterns, the following would look better
```suggestion
  name: {{ .Release.Name }}-st2web{{ template "enterpriseSuffix" . }}-ingress
```

erenatas: Oh, I did not realize there was such pattern. Will make my next commit perfect next time. Thanks a lot @armab !

PR #114, opened 2020-02-25
Allow injecting of hostAliases in the actionrunner deployment

AngryDeveloper:
This would allow the actionrunner to work with exotic DNS configuration
eg: you have access to some bastion host through its IP but you can't
resolve its name
The config directive is completely optional and won't impact existing
deployments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## In Development
 * Change ingress name from `<release name>-ingress` to <release name>-st2web-ingress, useful when using `stackstorm-ha` as a requirement for another chart. (#112) (by @erenatas)
 * Fix st2web ingress which should have been defined as an Integer instead of a String (#111) (by @erenatas)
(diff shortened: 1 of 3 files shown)


punkrokk:
Do we need to also have a docs PR?

armab:
@punkrokk Not for this change.

AngryDeveloper:
This should be better !

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## In Development
 * Change ingress name from `<release name>-ingress` to <release name>-st2web-ingress, useful when using `stackstorm-ha` as a requirement for another chart. (#112) (by @erenatas)
 * Fix st2web ingress which should have been defined as an Integer instead of a String (#111) (by @erenatas)
+* Add an option to inject hostAliases in the st2actionrunner containers
```
```suggestion
* Add an option to inject hostAliases in the st2actionrunner containers #114
```

Issue #9, opened 2018-09-20
Crypto key for st2 K/V datastore


armab:
> Moved from https://github.com/StackStorm/st2enterprise-dockerfiles/issues/52

In K8s objects and Helm chart we need to codify a way for user to specify custom crypto key for st2 datastore.
The secret key should be shared with corresponding stackstorm service as a file (`st2pi`, `st2actionrunners` etc).

See: https://docs.stackstorm.com/datastore.html#securing-secrets-admin-only

PR #115, opened 2020-02-26
Allow usage of datastore encryption with a pregenerated key

AngryDeveloper:
Closes https://github.com/StackStorm/stackstorm-ha/issues/9

This will allow the deployment to use datastore encryption if we provide
the required key through a K8S secret.

The secret's name and the key containing the effective json is configurable through values.yaml

By default the feature is disabled (and the chart behavior is unchanged).

The secret is mounted in the following deployments:

- st2api
- st2actionrunner
- st2sensorcontainers
- st2workflowengine
- ~st2rulesengine~

I might have missed some where the key is a dependency but I didn't run into any issue during my tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,6 @@
 * Add an option to inject hostAliases in the st2actionrunner containers (#114)
 * Add support for Service Accounts (#117) (by @Vince-Chenal)
 
-
 ## v0.24.0
 * Fix st2web ingress to use `/` path by default instead of `/*`, useful for nginx ingress controller (#103) (by @erenatas)
 * Add ability of templating on `st2.keyvalue` in Helm Values (#108) (by @erenatas)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Change ingress name from `<release name>-ingress` to <release name>-st2web-ingress, useful when using `stackstorm-ha` as a requirement for another chart. (#112) (by @erenatas)
 * Fix st2web ingress which should have been defined as an Integer instead of a String (#111) (by @erenatas)
 * Add an option to inject hostAliases in the st2actionrunner containers (#114)
+* Allow injection of datastore key in cluster (#115) (by @AngryDeveloper)
 
 ## v0.24.0
 * Fix st2web ingress to use `/` path by default instead of `/*`, useful for nginx ingress controller (#103) (by @erenatas)
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -453,9 +453,6 @@ spec:
         heritage: {{ .Release.Service }}
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmaps_st2-conf.yaml") . | sha256sum }}
-        {{- if $.Values.secrets.st2.datastore_crypto_key }}
-        checksum/datastore-key: {{ include (print $.Template.BasePath "/secrets_datastore_crypto_key.yaml") . | sha256sum }}
-        {{- end }}
     spec:
       {{- if .Values.enterprise.enabled }}
       imagePullSecrets:
@@ -478,11 +475,6 @@ spec:
         - name: st2-config-vol
           mountPath: /etc/st2/st2.user.conf
           subPath: st2.user.conf
-        {{- if .Values.secrets.st2.datastore_crypto_key }}
-        - name: st2-encryption-key-vol
-          mountPath: /etc/st2/keys
-          readOnly: true
-        {{- end }}
         resources:
 {{ toYaml .Values.st2rulesengine.resources | indent 10 }}
     {{- if .Values.st2rulesengine.serviceAccount.attach }}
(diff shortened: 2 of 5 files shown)


armab:
Closes https://github.com/StackStorm/stackstorm-ha/issues/9

values.yaml
```diff
@@ -62,6 +62,17 @@ st2:
...
+  # Optional configuration for handling a precomputed datastore encryption key
+  datastoreEncryption:
+    enabled: false
+    # Where will the key be mounted in the containers
+    encryptionKeyPath: /etc/st2/keys
+  # secret:
+  #   # The name of the secret in which the key is stored
+  #   name: st2-encryption-key
+  #   # The key containing the json formated encryption key
+  #   key: datastore_encryption_key.json
```
armab: For simplicity reasons, I'd prefer only the following Helm value to be exposed to the user:

```yaml
secrets:
  st2:
    # ST2 crypto key for the  K/V datastore. 
    # See https://docs.stackstorm.com/datastore.html#securing-secrets-admin-only for more info.
    # Warning! Replace with your own generated key!
    #datastore_crypto_key: {"hmacKey": {"hmacKeyString": "", "size": 256}, "size": 256, "aesKeyString": "", "mode": "CBC"}    
```

And so if `secrets.st2.datastore_crypto_key` value is present, st2 crypto key for the datastore will be enabled automatically with all the K8s secrets under the hood and defaults like https://github.com/StackStorm/st2-packages/blob/e86adb5e75b5013dac43198f9e464b29cb08ed58/scripts/st2bootstrap-deb.sh#L335-L361


See for example other secrets like https://github.com/StackStorm/stackstorm-ha/blob/master/templates/secrets_ssh.yaml for more mechanics.

With that, crypto key path would be `/etc/st2/keys/datastore_key.json`.
Services that need shared access are `st2api st2sensorcontainer st2workflowengine st2actionrunner` and `st2client` pods.

AngryDeveloper: I've implemented as you advised, it's way less complex for the end user indeed.

AngryDeveloper:
> 
> 
> We'll also need checksum annotations referencing new secret.
> 
> See
> 
> https://github.com/StackStorm/stackstorm-ha/blob/d59b8c4d9ea133fa0283c975b0053360381c153e/templates/deployments.yaml#L998-L1000
> 
> for example

Will implement as soon as possible.

templates/deployments.yaml
```diff
@@ -598,12 +633,25 @@ spec:
...
         - name: st2-config-vol
           configMap:
             name: {{ .Release.Name }}-st2-config
+        {{- if .Values.secrets.st2.datastore_crypto_key }}
+        - name: st2-encryption-key-vol
+          secret:
+            secretName: {{ .Release.Name }}-st2-datastore-crypto-key
+            items:
+            - key: datastore_crypto_key
+              path: encryption_key.json
```
armab:
```suggestion
              path: datastore_key.json
```

templates/deployments.yaml
```diff
@@ -447,12 +466,25 @@ spec:
         - name: st2-config-vol
           mountPath: /etc/st2/st2.user.conf
           subPath: st2.user.conf
+        {{- if .Values.secrets.st2.datastore_crypto_key }}
```
armab: Following the https://github.com/StackStorm/st2-packages/blob/e86adb5e75b5013dac43198f9e464b29cb08ed58/scripts/st2bootstrap-deb.sh#L356-L360 I don't think `st2rulesengine` needs access to st2 datastore crypto key.

@m4dcoder can you confirm?

templates/deployments.yaml
```diff
@@ -831,6 +879,8 @@ spec:
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmaps_st2-conf.yaml") $ | sha256sum }}
         checksum/packs: {{ include (print $.Template.BasePath "/configmaps_packs.yaml") $ | sha256sum }}
+        {{- if $.Values.secrets.st2.datastore_crypto_key }}
+        checksum/datastore-key: {{ include (print $.Template.BasePath "/secrets_datastore_crypto_key.yaml") & | sha256sum }}
```
armab, line 883: `Helm-lint` is unhappy about this one: https://app.circleci.com/pipelines/github/StackStorm/stackstorm-ha/570/workflows/86ddf92e-4929-4456-8cac-0c76d0085ee0/jobs/2231/steps

```suggestion
        checksum/datastore-key: {{ include (print $.Template.BasePath "/secrets_datastore_crypto_key.yaml") $ | sha256sum }}
```
Looks like it's a typo here.

templates/deployments.yaml
armab: `st2rulesengine` doesn't need access to st2 datastory crypto key. Please remove it.

CHANGELOG.md
```diff
@@ -27,6 +28,7 @@
 * Add an option to inject hostAliases in the st2actionrunner containers (#114)
 * Add support for Service Accounts (#117) (by @Vince-Chenal)
 
+
```
armab, line 31: To minimize the diff
```suggestion
```

templates/deployments.yaml
```diff
@@ -831,6 +871,8 @@ spec:
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmaps_st2-conf.yaml") $ | sha256sum }}
         checksum/packs: {{ include (print $.Template.BasePath "/configmaps_packs.yaml") $ | sha256sum }}
+        {{- if $.Values.secrets.st2.datastore_crypto_key }}
```
armab, line 874: missing a closing `end` tag for `if`?

templates/deployments.yaml
```diff
@@ -979,6 +1034,8 @@ spec:
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmaps_st2-conf.yaml") . | sha256sum }}
         checksum/ssh: {{ include (print $.Template.BasePath "/secrets_ssh.yaml") . | sha256sum }}
+        {{- if $.Values.secrets.st2.datastore_crypto_key }}
```
armab, line 1037: same with the closing `if`

templates/deployments.yaml
armab: In fact, do we really need `if` -> `end` for the datastore crypto key checksum in annotations?
I mean even if `secrets_datastore_crypto_key.yaml` is empty if no key is provided, when it's populated the checksum is changed.

AngryDeveloper: Yep, not quite awake yet...

Indeed, removing the conditional altogether would be cleaner, implementing.

PR #117, opened 2020-03-05
Add Service Account Management

Vince-Chenal:
The goal of this PR is to:
- allow creating a Service Account dedicated to st2 from the Chart
- attach it (or another external one) to the pods you want

On my side, I need this to give st2actionrunner access to AWS via a Service Account.
https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,7 @@
 * Change ingress name from `<release name>-ingress` to <release name>-st2web-ingress, useful when using `stackstorm-ha` as a requirement for another chart. (#112) (by @erenatas)
 * Fix st2web ingress which should have been defined as an Integer instead of a String (#111) (by @erenatas)
 * Add an option to inject hostAliases in the st2actionrunner containers (#114)
-
-## v0.25.0
-* Add support for Service Accounts (by @Vince-Chenal)
+* Add support for Service Accounts (#117) (by @Vince-Chenal)
 
 ## v0.24.0
 * Fix st2web ingress to use `/` path by default instead of `/*`, useful for nginx ingress controller (#103) (by @erenatas)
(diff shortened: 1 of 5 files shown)


values.yaml
```diff
@@ -53,6 +53,17 @@ enterprise:
...
+##
+## Service Account
+##
+serviceAccount:
+  # Whether the Chart should create the service account or not
+  create: true
+  # Used to define service account annotations
+  serviceAccountAnnotations: {}
+  # Used to override service account name
+  serviceAccountName:
```
mickmcgrath13: Why not put this per service rather than a centrally located serviceAccount?

For example, what if we wanted st2stream to use a different serviceaccount from st2api?

armab: Following the other official Helm charts and templating examples from the https://github.com/helm/charts/search?q=serviceaccount&unscoped_q=serviceaccount looks like it's pretty good default to start with. There is however a good opportunity to extend the base functionality in future, if there will be advanced use cases you mentioned.

Vince-Chenal: Hello,
sorry for the delay.
I added a new version in the CHANGELOG, is this what I was supposed to do ?

Vince-Chenal:
Hello,
sorry for the delay.
I added a new version in the CHANGELOG, is this ok  ?

values.yaml
Vince-Chenal: Hello @mickmcgrath13,
I asked myself the question but it felt like a good start.
Personally I only need a service account and I did not have a use case in mind for having multiple service accounts.

But we can easily iterate over this first step.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 * Fix st2web ingress which should have been defined as an Integer instead of a String (#111) (by @erenatas)
 * Add an option to inject hostAliases in the st2actionrunner containers (#114)
 
+## v0.25.0
```
```suggestion
```

The changes from this PR should go into `In Development` section.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
...
+* Add support for Service Accounts (by @Vince-Chenal)
```
```suggestion
* Add support for Service Accounts (#117) (by @Vince-Chenal)
```

PR #118, opened 2020-03-09
Allow mounting of NFS volumes instead of using st2packs for pack management

### LINKED ISSUE #18: Sharing pack content: Bundle custom st2packs image vs Shared NFS storage?
### Bundling custom st2 packs as immutable Docker image
In current implementation for this chart to use custom st2 packs we rely on building a dedicated Docker image with pack content and virtualenv, pre-installed and bundled beforehand, see: https://docs.stackstorm.com/latest/install/ewc_ha.html#custom-st2-packs
As a downside means any writes like `st2 pack install` or saving the workflow from `st2flow` won't work in current HA env.

### Share content via ReadWriteMany shared file system
There is an alternative approach, - sharing pack content via read-write-many NFS (Network File System) as [High Availability Deployment doc](https://docs.stackstorm.com/latest/reference/ha.html) recommends. 

### Examples
For example, There is a stable Helm chart [`nfs-server-provisioner`](https://github.com/helm/charts/tree/master/stable/nfs-server-provisioner) which codifies NFS in an automated way.

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes
lists K8s supported solutions for `ReadWriteMany` (RWX) volumes.

From that list of volumes, additional interest apart of NFS, goes to `CephFS` and https://github.com/rook/rook `Rook`, a CNCF-hosted project for storage orchestration.

### Feedback Needed!

As beta is in progress and both methods have their pros and cons, we’d like to hear your feedback, ideas, experience running any of these shared file-systems and which way would work better for you.

AngryDeveloper:
This should close #18

This will allow the end users to mount NFS volumes to manage their custom packs. To be noted: when using an NFS mount, the lifecycle management of the packs (installing/updating/registering) is deferred back to the stackstorm administrators.
One way to do it would be to trigger packs installation and/or registration through the API using
https://api.stackstorm.com/api/v1/packs/#/packs_controller.install.post
https://api.stackstorm.com/api/v1/packs/#/packs_controller.register.post
or directly through the st2client pod.

The st2packs method is of course still available. If both `st2.packs.image.repository` and `st2.packs.nfs.server` are set, the st2packs image will be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Change ingress name from `<release name>-ingress` to <release name>-st2web-ingress, useful when using `stackstorm-ha` as a requirement for another chart. (#112) (by @erenatas)
 * Fix st2web ingress which should have been defined as an Integer instead of a String (#111) (by @erenatas)
 * Add an option to inject hostAliases in the st2actionrunner containers (#114)
+* Add an option to mount NFS volumes instead of using the `st2packs` image (#18) (by @AngryDevelopper)
 
 ## v0.24.0
 * Fix st2web ingress to use `/` path by default instead of `/*`, useful for nginx ingress controller (#103) (by @erenatas)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Change ingress name from `<release name>-ingress` to <release name>-st2web-ingress, useful when using `stackstorm-ha` as a requirement for another chart. (#112) (by @erenatas)
 * Fix st2web ingress which should have been defined as an Integer instead of a String (#111) (by @erenatas)
 * Add an option to inject hostAliases in the st2actionrunner containers (#114)
-* Add an option to mount NFS volumes instead of using the `st2packs` image (#18) (by @AngryDevelopper)
+* Add an option to mount NFS volumes instead of using the `st2packs` image (#118) (by @AngryDevelopper)
 
 ## v0.24.0
 * Fix st2web ingress to use `/` path by default instead of `/*`, useful for nginx ingress controller (#103) (by @erenatas)
--- a/templates/jobs.yaml
+++ b/templates/jobs.yaml
@@ -323,8 +323,8 @@ spec:
       {{- if .Values.st2.packs.image.pullSecret }}
       - name: {{ .Values.st2.packs.image.pullSecret }}
       {{- end }}
-      {{- if .Values.st2.packs.image.repository }}
       initContainers:
+      {{- if .Values.st2.packs.image.repository }}
       # Merge packs and virtualenvs from st2actionrunner with those from the st2.packs image
       # Custom packs
       - name: st2-custom-packs
@@ -341,6 +341,7 @@ spec:
           - |
             /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
             /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+      {{ end }}
       # System packs
       - name: st2-system-packs
         image: "{{ template "imageRepository" . }}/st2actionrunner{{ template "enterpriseSuffix" . }}:{{ .Chart.AppVersion }}"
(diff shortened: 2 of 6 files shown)


armab:
@hnanchahal I think you might be interested in this story of sharing pack content via NFS mount as you did something similar.
Can you also take a look at PR and provide feedback?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Change ingress name from `<release name>-ingress` to <release name>-st2web-ingress, useful when using `stackstorm-ha` as a requirement for another chart. (#112) (by @erenatas)
 * Fix st2web ingress which should have been defined as an Integer instead of a String (#111) (by @erenatas)
 * Add an option to inject hostAliases in the st2actionrunner containers (#114)
+* Add an option to mount NFS volumes instead of using the `st2packs` image (#18) (by @AngryDevelopper)
```
```suggestion
* Add an option to mount NFS volumes instead of using the `st2packs` image (#118) (by @AngryDevelopper)
```

values.yaml
```diff
@@ -86,6 +86,18 @@ st2:
       # Optional name of the imagePullSecret if your custom packs image is hosted by a private Docker registry behind the auth
       #pullSecret: st2packs-auth
 
+    # Custom packs and virtualenvs can be mounted using NFS.
```
armab: Great start! :+1: 

I think we'll need a more user-friendly story in both README.md and then after merging this PR at https://docs.stackstorm.com/install/k8s_ha.html which would instruct how to install/use this as alternative to `st2pack` image.

AngryDeveloper:
> Also can you provide an example about how to install/configure NFS in K8s and respective Helm values? So someone reviewing this will follow that to verify this feature/story.

Isn't that a bit out of scope ?

armab:
Having the full example is out of scope for the README, but it'll be good to have this kind of info in this issue to help someone test and try it end to end.

armab:
Overall with NFS like distributing pack content we provide a story/solution for the user. 
For instance, custom st2packs has
https://docs.stackstorm.com/install/k8s_ha.html#custom-st2-packs
https://github.com/stackstorm/stackstorm-ha#install-custom-st2-packs-in-the-cluster
https://github.com/StackStorm/st2packs-dockerfiles/
as a helpers to guide users about preparing custom packs in K8s environment.
Alternative solution with NFS shouldn't have less information about how to use it and overall what steps to follow.

templates/jobs.yaml
```diff
@@ -342,6 +341,8 @@ spec:
           - |
             /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
             /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+      {{ end }}
+      {{- if (.Values.st2.packs.image.repository) or (.Values.st2.packs.nfs.server) }}
```
armab: Default system packs should be always installed and registered, irrespective of the custom packs.

AngryDeveloper: https://github.com/StackStorm/stackstorm-ha/blob/85c21ed9e925b6fabcbcdef6eae3e5ee04cd8b34/templates/jobs.yaml#L327-L358

On master it seems that if `st2.packs.image.repository` is not set the whole `initContainers` section of the job is omitted, I tried to mirror this behavior, I might have made a mistake though.

armab: That's right, confirming. I misread the diff.

The default pack content is already present in st2actionrunner image and is registered if no custom packs are provided.

values.yaml
AngryDeveloper: I've added more details to the README, when everything is squared on stackstorm-ha repo I'll take a look into updating the main stackstorm doc.

chris3081:
I would like to see a deployment engine in front of this like a Jenkins lite (API / UI) which has write on the volume.. It would offer a clean deployment solution and might allow us to have something in the runners, st2api, st2client etc to install pack OS dependencies, tho I'm not 100% sure how it would work at the moment, but maybe the actionrunner could be commanded by the deployment engine to install OS packs (would need to account for the st2api which runs as stanley)

ericreeves:
Greetings!  This PR needs a rebase given the big upstream changes from master, which I have been unsuccessful at so far.  Anybody else have any luck?  :)

cognifloyd:
Thanks for this PR! I used some of the ideas here in building #199 . I need flexVolume + ceph-rook, you need NFS, and others will need who knows which storage solution, so I tried to make something that was both generic and simple. What do you think of #199? Will that work for you?

cognifloyd:
@AngryDeveloper could you review #199? Is there anything missing from #199 that you included in this PR, or that you think would be better a different way?

cognifloyd:
If #199 is satisfactory for you, then I think we could close this PR. What do you think?

Issue #14, opened 2018-09-20
Rearrange "secrets" in Helm values


armab:
Currently in Helm `values.yaml` for this chart there is a dedicated `secrets` section holding values for data like st2 username/password, SSL certs, SSH keys and so on.

~We wanted to move it to dedicated `secrets.yaml` when it gets implemented in Helm (https://github.com/kubernetes/helm/issues/2196).~ (now closed)

**Update:** We'll need to just obsolete dedicated `secrets` block and merge it with all other sections as any official Helm chart does without dedicated sub-section weirdness.

https://github.com/StackStorm/stackstorm-enterprise-ha/blob/95fe6bb5f0d76ef772ad3f4391ff5cd3e00d1d86/values.yaml#L82-L100

~Feedback needed: which approach/way looks better and friendlier to you?~

mickmcgrath13:
Why not separate out the secrets to their respective services rather than aggregating them.  
i.e.
```
serviceA:
  secrets:
    foo: a
serviceB:
  secrets:
    bar: b
```
instead of:
```
secrets:
  serviceA:
    foo: a
  serviceB:
    bar: b
```
?

This would still work well with a separate `secrets.yaml` file

armab:
It all depends and on what's the future of https://github.com/helm/helm/issues/2196 and where it'll go or whether it'll get implemented anytime.

Looking back, feels like going with YOLO approach and planning for "now":

```
st2:
  username: st2admin
  password: ***
```
what's used by the other official Helm charts was a more natural way comparing to:
```
secrets:
  st2:
    username: st2admin
    password: ***
```
or
```
st2:
  secrets:
    username: st2admin
    password: ***
```

AngryDeveloper:
In any case, relying on helm/helm#2196 is a long shot, the issue wasn't very active lately. I don't have any issues using either of the solution you guys mentioned, as long as we can progress on the features that we want to implement :)

armab:
https://github.com/helm/helm/issues/2196 was just closed. I agree by the maintainer, considering it wasn't addressed by anyone until this moment.

Hence we should just get rid of using `secrets` keyword for Helm values and rearrange them.

PR #120, opened 2020-03-10
Allow injection of env vars in any deployment

### LINKED ISSUE #119: st2chatops env/secrets management
Hey there !

Currently the injection of secret in `st2chatops` is done through an env variable mapping.
The env is iterated over as follow: https://github.com/StackStorm/stackstorm-ha/blob/85c21ed9e925b6fabcbcdef6eae3e5ee04cd8b34/templates/secrets_st2chatops.yaml#L18-L21

This seems to be a bit counter intuitive. From the end user point of view, the env is never referenced as being secret, but every K\V is stored as a secret in k8s.

In addition to this, if we wanted to inject custom env variables that don't have anything secret (eg: proxy settings or equivalent) they will forcefully be stored as secret.

I ran into this issue when trying to allow injection of arbitrary env variables into the deployments.

One solution would be using `secrets.st2` for the variables that should be stored as secret, and a classic `st2chatops.env` or more generally `(st2chatops|st2actionrunner|st2api|...).env` variable to store the "standard" environment.
This can be loosely linked to #14

AngryDeveloper:
This allows the user to inject any custom env variable in any deployment. 

The need was the ability to configure a proxy so that `st2 pack install` doesn't fail when behind said proxy. The env variable needed to be injected in the `st2actionrunner` deployments.

There is a BC break on `st2chatops`  configuration, if secrets are provided, they need to be added to `secrets.st2.st2chatops.env` instead of `st2chatops.env`.

This should also closes #119 and partially solves #14.


values.yaml
```diff
@@ -158,6 +160,9 @@ ingress:
 # TODO: Alternatively as part of reorganizing Helm values, consider moving values to existing `st2` and `st2web` sections ? (#14)
 secrets:
   st2:
+    st2chatops:
```
mickmcgrath13: Any reason this is `secrets.st2.st2chatops` instead of `secrets.st2chatops` (i.e. why the nesting under `st2`)?

mickmcgrath13: Also, any reason not to put the secrets specific to st2chatops under st2chatops (and the same for all other services) instead of aggregating all secrets to a central place?

i.e. instead of 
```
secrets:
  st2:
    st2chatops:
      foo: bar
```
do
```
st2chatops:
  secrets:
    foo: bar
```

AngryDeveloper: > Any reason this is `secrets.st2.st2chatops` instead of `secrets.st2chatops` (i.e. why the nesting under `st2`)?

For this part, not particular reason, it's true that it's a bit too verbose.

> Also, any reason not to put the secrets specific to st2chatops under st2chatops (and the same for all other services) instead of aggregating all secrets to a central place?

See issue #14 :)

mickmcgrath13: Makes sense.  I think it'd be preferable to spread the secrets out into their respective services, but we can tackle that when we tackle #14 

Thanks!

armab: I agree that `secrets.st2.st2chatops` looks a bit too verbose and even counter-intuitive.

This also raises some inconsistencies in the context of https://github.com/StackStorm/stackstorm-ha/issues/119 discussion. I'm thinking about exploring the alternatives for the Helm values definition.

templates/secrets_st2chatops.yaml
```diff
@@ -16,7 +16,7 @@ metadata:
     heritage: {{ .Release.Service }}
 type: Opaque
 data:
-{{- range $env, $value := .Values.st2chatops.env }}
+{{- range $env, $value := .Values.secrets.st2.st2chatops.env }}
```
armab, line 19: I think there is at least an agreement about the :100: need of raw `ENV` for every Pod and that's super helpful.
Can we keep raw ENV functionality in this PR to be able to merge it asap, but split the `st2chatops` secrets ENV enhancement into another PR?

templates/deployments.yaml
```diff
@@ -70,6 +70,13 @@ spec:
         # TODO: Add liveness/readiness probes (#3)
         #livenessProbe:
         #readinessProbe:
+        {{- if .Values.st2auth.env }}
+        env:
+        {{- range $env, $value := .Values.st2auth.env }}
+        - name: {{ $env | quote }}
+          value: {{ $value | quote }}
+        {{- end }}
+        {{- end }}
```
cognifloyd, line 79: I think this could be extracted into a helper to avoid some repetition.
So, here it would have:
```
{{ include "custom-env" .Values.st2auth | indent 8 }}
```
And then the template would be:
```
{{- define "custom-env" -}}
  {{- if .env }}
env:
    {{- range $env, $value := .env }}
- name: {{ $env | quote }}
  value: {{ $value | quote }}
    {{- end }}
  {{- end }}
{{- end -}}
```

templates/deployments.yaml
```diff
@@ -866,6 +929,13 @@ spec:
           - --single-sensor-mode
           - --sensor-ref={{ .ref }}
         {{- end }}
+        {{- if .env }}
+        env:
+        {{- range $env, $value := .env }}
+        - name: {{ $env | quote }}
+          value: {{ $value | quote }}
+        {{- end }}
+        {{- end }}
```
cognifloyd, line 938: And the template here would just be:
```
{{ include "custom-env" . | indent 8 }}
```

templates/deployments.yaml
```diff
@@ -1369,6 +1457,10 @@ spec:
             secretKeyRef:
               name: {{ .Release.Name }}-st2-auth
               key: password
+{{- range $env, $value := .Values.st2chatops.env }}
+        - name: {{ $env | quote }}
+          value: {{ $value | quote }}
+{{- end }}
```
cognifloyd, line 1463: I think I would drop this, and leave the st2chatops env as is (in `envFrom[1].secretRef` below).
```suggestion
```

templates/deployments.yaml
```diff
@@ -1242,6 +1326,10 @@ spec:
         env:
         - name: ST2CLIENT
           value: "1"
+{{- range $env, $value := .Values.st2client.env }}
+        - name: {{ $env | quote }}
+          value: {{ $value | quote }}
+{{- end }}
```
cognifloyd, line 1332: A couple of jobs also have an `env:` block already. So, we could split the template into two to accommodate this usage.
```
{{- define "custom-env-entries" -}}
  {{- range $env, $value := .env }}
- name: {{ $env | quote }}
  value: {{ $value | quote }}
  {{- end }}
{{- end -}}
{{- define "custom-env" -}}
  {{- if .env }}
env:
{{ include "custom-env-entries" . }}
  {{- end }}
{{- end -}}
```
Which would make this block be:
```
{{ include "custom-env-entries" .Values.st2client }}
```

templates/secrets_st2chatops.yaml
cognifloyd, line 19: Yeah, could you revert this? The `.Values.secrets` stuff needs an overhaul in a separate PR.
```suggestion
{{- range $env, $value := .Values.st2chatops.env }}
```

values.yaml
```diff
@@ -158,6 +160,9 @@ ingress:
...
+      env:
+        HUBOT_SLACK_TOKEN: xoxb-CHANGE-ME-PLEASE
```
cognifloyd, line 165:
```suggestion
```

values.yaml
```diff
@@ -103,6 +103,8 @@ st2:
         affinity: {}
         nodeSelector: {}
         tolerations: []
+        env: {}
+        # HTTP_PROXY: http://proxy:1234
```
cognifloyd, line 107: Nice. I like the commented example. :+1:

values.yaml
```diff
@@ -348,7 +379,6 @@ st2chatops:
   # for the full list of supported adapters and example ENV variables.
   env:
     HUBOT_ADAPTER: slack
-    HUBOT_SLACK_TOKEN: xoxb-CHANGE-ME-PLEASE
```
cognifloyd, line 351: Revert this too please.

cognifloyd (changes requested):
Awesome start!

I think we might also want to add an env block to the jobs (all the jobs would share the same block).
In `values.yaml`, it would be something like:
```
jobs:
  env: {}
```

templates/secrets_st2chatops.yaml
cognifloyd: Following up: `.Values.secrets` was removed in #203

cognifloyd:
I want to clear out old PRs by merging them if possible.
So, I just merged in master and made a few updates to address all the comments here.

I think a Changelog entry is the only thing left before merging this. @AngryDeveloper what do you think of these changes?

armab (approved):
The changes look really helpful! :+1: 

Thanks @cognifloyd. I think it's also OK to inject the changelog as well to merge it.

cognifloyd (approved):

PR #123, opened 2020-04-02
Added custom annotations to sensorcontainer and actionrunner Pods

### LINKED ISSUE #122: Add custom annotations to Sensorcontainer and Actionrunner Pods
For example, if there are actions to be performed in AWS or events to be retrieved from AWS, the `actionrunner` Pods or the `sensorcontainer` Pods require a IAM role attached to them, in annotations:

`iam.amazonaws.com/role: <Role_ARN>`

This is just a use case for adding a custom annotation to one or more Pods. I think the chart should support such behaviour.

stefangusa:
Closes #122

### DIFF
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -828,9 +828,12 @@ spec:
         chart: {{ $.Chart.Name }}-{{ $.Chart.Version }}
         release: {{ $.Release.Name }}
         heritage: {{ $.Release.Service }}
+    {{- with .annotations }}
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmaps_st2-conf.yaml") $ | sha256sum }}
         checksum/packs: {{ include (print $.Template.BasePath "/configmaps_packs.yaml") $ | sha256sum }}
+{{ toYaml . | indent 8 }}
+    {{- end }}
     spec:
       imagePullSecrets:
       {{- if $.Values.enterprise.enabled }}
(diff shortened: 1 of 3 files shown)


templates/deployments.yaml
```diff
@@ -973,9 +976,12 @@ spec:
         chart: {{ .Chart.Name }}-{{ .Chart.Version }}
         release: {{ .Release.Name }}
         heritage: {{ .Release.Service }}
+      {{- with .Values.st2actionrunner.annotations }}
```
armab, line 979: By using `with` here the bracket will hide the required `checksum/config` and `checksum/ssh` annotations if st2actionrunner annotations is empty.

stefangusa: Solved

templates/deployments.yaml
```diff
@@ -976,6 +981,11 @@ spec:
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmaps_st2-conf.yaml") . | sha256sum }}
         checksum/ssh: {{ include (print $.Template.BasePath "/secrets_ssh.yaml") . | sha256sum }}
+        {{- if .Values.st2actionrunner.annotations }}
+        {{- range $key, $value := .Values.st2actionrunner.annotations }}
+        {{ $key }}: {{ $value | quote }}
+        {{- end }}
```
armab, line 987: Is there any advantage of iterating comparing to more simple construct?

Can we use just use
```suggestion
{{ toYaml .Values.st2actionrunner.annotations | indent 8 }}
```
?

stefangusa: Definitely we should use what you propose, I was just not sure it works (I looked for a similar construct and I only encountered the iteration approach). Thank you!

Issue #94, opened 2019-10-23
'stable/etcd-operator' is not really stable for [coordination]


armab:
After seeing the following in logs when cluster couldn't start itself or even start clean if all `etcd` pods were killed:
```
level=warning msg="all etcd pods are dead." cluster-name=etcd-cluster cluster-namespace=default pkg=cluster
```
This situation is not recovered by `etcd-operator`.
https://github.com/coreos/etcd-operator/blob/8347d27afa18b6c76d4a8bb85ad56a2e60927018/pkg/cluster/cluster.go#L248-L252

Researching further looks like there are quite a lot of cases when `etcd-operator` can't recover itself:
- https://github.com/coreos/etcd-operator/issues/1973
- https://github.com/coreos/etcd-operator/issues/1559
- https://github.com/coreos/etcd-operator/issues/1972
- https://github.com/coreos/etcd-operator/issues/2044

----

Because this backend is needed just for short-lived coordination locks, consider switching to `Redis` or even single-instance `etcd` like it was before (https://github.com/StackStorm/stackstorm-ha/pull/52)?

trstruth:
:(

danielburrell:
Further to this, the etc-operator project has just been archived and is now in maintenance mode. Not good to have unfinished components slowly rot over time. Can we drop the etcd-operator dependency somehow? 

I wrote something here earlier
https://forum.stackstorm.com/t/etcd-operator-project-archived/1140

armab:
@danielburrell Thanks for letting us know about etc-operator archived state.

Yes, the next step would be identifying another coordination backend that has good helm charts and works well.
It could be Redis or Memcached or other alternatives from https://docs.openstack.org/tooz/latest/user/drivers.html

PR #127, opened 2020-04-29
Added toggling of etcd-operator as a coordination backend

rrahman-nv:
Added toggling of etcd-operator as a coordination backend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## In Development
+* Added support to toggle etcd-operator as a coordination backend (#127) (by @rrahman-nv)
 
 ## v0.26.0
 * Added custom annotations to sensorcontainer and actionrunner Pods (#123) (by @stefangusa)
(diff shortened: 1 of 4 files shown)


armab:
Somewhat related to https://github.com/StackStorm/stackstorm-ha/issues/94 and https://github.com/StackStorm/st2-dockerfiles/pull/25

And so this PR allows someone to disable etcd-operator deployment in K8s cluster and use their own Redis instance for coordination backend. cc @trstruth you might be interested. 
This is somewhat similar to using custom RabbitMQ and MongoDB clusters outside of K8s deployment.

wantdrink:
Wonderful PR. Thanks a lot @rrahman-nv  & @armab , that really helps.
A simple question: can I configure the redis 5.0 with sentinels and configure the cluster-like URL in st2.conf?

armab:
Sure, that could be configured.
We're using tooz library for st2 `[coordination]` which abstracts common functionality for different backends.

For Redis cluster connection string examples see:
https://github.com/openstack/tooz/blob/master/tooz/drivers/redis.py

wantdrink:
@armab  awesome! Thanks a lot.

PR #131, opened 2020-05-11
provide option for alternate docker repository

ytjohn:
We need the option to pull images from an internal docker repository. This PR allows us to specify the name of the repository, while still allowing default and enterprise.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,9 @@ jobs:
           name: Helm Lint Check (Enterprise)
           command: helm lint --set enterprise.enabled=true --set enterprise.license=123asd456fake
       - run:
+          name: Helm Lint Check (st2Repository)
+          command: helm lint --set image.st2Repository="docker.example.com/st2modified"
+      - run:
           name: Helm template
           command: |
             mkdir -p enterprise community
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,9 +33,6 @@ jobs:
           name: Helm Lint Check (Enterprise)
           command: helm lint --set enterprise.enabled=true --set enterprise.license=123asd456fake
       - run:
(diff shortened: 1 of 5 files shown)


values.yaml
```diff
@@ -7,6 +7,12 @@
 image:
   # Image pull policy. Change to "IfNotPresent" when switching to stable images
   pullPolicy: Always
+  # st2 image repository. Set this to override the default ("stackstorm") or enterprise
+  # docker image repository ("docker.stackstorm.com"). Applies to all st2 containers except
+  # st2chatops and st2packs (which have their own override). This also does not impact 
+  # dependencies such as mongo or redis, which have their own helm chart settings.
+  st2Repository: ""
```
armab, line 14:
```suggestion
  repository: ""
```
Would be better to follow the consistent naming as in other Helm values.

`image.repository`

.circleci/config.yml
```diff
@@ -32,6 +32,9 @@ jobs:
       - run:
           name: Helm Lint Check (Enterprise)
           command: helm lint --set enterprise.enabled=true --set enterprise.license=123asd456fake
+      - run:
+          name: Helm Lint Check (st2Repository)
+          command: helm lint --set image.st2Repository="docker.example.com/st2modified"
```
armab, line 37: It's a bit excessive to have a dedicated lint check for this small thing.
I think this would better fit into #28 Unit Tests which (we don't have yet).

values.yaml
ytjohn: changed

.circleci/config.yml
ytjohn: removed

armab:
This change was released as `v0.28.0` chart version:
https://github.com/StackStorm/stackstorm-ha/releases/tag/v0.28.0

Issue #11, opened 2018-09-20
Pod scheduling priority: Helm release could be stuck in a restart loop due Mongo/Rabbit clusters glued with a "Pending" state


armab:
> Moved from https://github.com/StackStorm/st2enterprise-dockerfiles/issues/80

Sometimes when deploying StackStorm cluster in K8s (especially with resource pressure), entire Helm deployment is stuck in a dead dependency loop: st2 pods keep restarting, rescheduling, recreating due to no MQ/DB connection, while Mongo and Rabbit cluster can't start because of st2 pods reschedule spam.

> Note that our HA deployment with 3 nodes for each DB and MQ, 2 for each st2 service creates a cluster of minimum `30` Pods

The solution could be trying K8s Pod priority https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/ which is beta starting from K8s `v1.11`.

This way we can prioritize scheduling for MongoDB, RabbitMQ and etcd clusters before st2 services.

PR #158, opened 2020-11-20
Updated RabbitMQ-Chart to 1.46.1 & improved Reboot-Resilience

moonrail:
Hello altogether

Potentially partly helps with the https://github.com/StackStorm/stackstorm-ha/issues/11

This Pull Request updates RabbitMQs-Chart-Dependency to 1.46.1, as on our kubernetes installations (1.17, 1.18, 1.19) 1.44.1 would not run due to mysterious disk-space complaints while being correctly assigned to a PersistentVolume and being able to RW to it.

As this new RabbitMQ-Version enables Prometheus-Monitoring by default, most of current installations would fail, therefore this is disabled it by default.

Enabled `rabbitmqErlangCookie`, as otherwise Cluster-Data is not reusable after RabbitMQ-Deployment-Restart/-Rebuild (or short period of 0 Replicas).

Added `forceBoot`, as otherwise Cluster-Data is not reusable after RabbitMQ-Deployment-Restart/-Rebuild (or short period of 0 Replicas), as Mnesia Tables are not cleaned up and cause RabbitMQ to not boot up. See https://github.com/helm/charts/issues/13485

So this should improve User Experience by not having to ditch PersistentVolumes after RabbitMQ-Redeployments.
<br>Not really sure about this - but if StackStorm holds queued Executions in RabbitMQ, this would also help in disaster cases, to not loose all running Executions.

Tested with 3.3dev on kubernetes 1.17, 1.18 & 1.19.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## In Development
 * Change pullPolicy to "IfNotPresent", as Docker-Hub Ratelimits now (#159) (by @moonrail)
-
+* Update `rabbitmq-ha` 3rd party chart from `1.44.1` to `1.46.1` (#158) (by @moonrail)
+* Enable `rabbitmqErlangCookie` for `rabbitmq-ha` by default, to ensure cluster-redeployments do not fail (#158) (by @moonrail)
+* Add `forceBoot` for `rabbitmq-ha` by default, to ensure cluster-redeployments do not fail due to unclean exits (#158) (by @moonrail)
 
 ## v0.32.0
 * Fix a bug when datastore encrypted keys didn't work in scheduled rules. datastore_crypto_key is now shared with the ``st2scheduler`` pods (#148) (by @rahulshinde26)
--- a/values.yaml
+++ b/values.yaml
@@ -450,6 +450,9 @@ rabbitmq-ha:
   # Change to `false` to disable in-cluster rabbitmq deployment.
   # Specify your external [messaging] connection parameters under st2.config
   enabled: true
+  # On unclean cluster restarts forceBoot is required to cleanup Mnesia tables (see: https://github.com/helm/charts/issues/13485)
+  # Use it only if you prefer availability over integrity.
+  forceBoot: true
   rabbitmqUsername: admin
   # TODO: Use default random 24 character password, but need to fetch this string for use by downstream services
   rabbitmqPassword: 9jS+w1u07NbHtZke1m+jW4Cj
@@ -458,7 +461,8 @@ rabbitmq-ha:
   #rabbitmqMemoryHighWatermark: 512MB
   #rabbitmqMemoryHighWatermarkType: absolute
   # Up to 255 character string, should be fixed so that re-deploying the chart does not fail (see: https://github.com/helm/charts/issues/12371)
-  #rabbitmqErlangCookie: 8MrqQdCQ6AQ8U3MacSubHE5RqkSfvNaRHzvxuFcG
+  # NB! It's highly recommended to change the default insecure rabbitmqErlangCookie value!
+  rabbitmqErlangCookie: 8MrqQdCQ6AQ8U3MacSubHE5RqkSfvNaRHzvxuFcG
   persistentVolume:
     enabled: true
   # RabbitMQ application vhost, should match with 'ha' Queue Mirroring definition policy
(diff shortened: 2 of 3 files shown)


values.yaml
```diff
@@ -450,6 +450,8 @@ rabbitmq-ha:
   # Change to `false` to disable in-cluster rabbitmq deployment.
   # Specify your external [messaging] connection parameters under st2.config
   enabled: true
+  # On unclean cluster restarts forceBoot is required to cleanup Mnesia tables (see: https://github.com/helm/charts/issues/13485)
+  forceBoot: true
```
armab, line 454: As every option is about trade-off, worth adding comment for this new default:

```suggestion
  # On unclean cluster restarts forceBoot is required to cleanup Mnesia tables (see: https://github.com/helm/charts/issues/13485)
  # Use it only if you prefer availability over integrity.
  forceBoot: true
```

values.yaml
```diff
@@ -458,7 +460,7 @@ rabbitmq-ha:
   #rabbitmqMemoryHighWatermark: 512MB
   #rabbitmqMemoryHighWatermarkType: absolute
   # Up to 255 character string, should be fixed so that re-deploying the chart does not fail (see: https://github.com/helm/charts/issues/12371)
```
armab, line 462: As we're defaulting `rabbitmqErlangCookie` value for everyone, let's at least include a warning comment as recommendation to change the default.

```suggestion
  # Up to 255 character string, should be fixed so that re-deploying the chart does not fail (see: https://github.com/helm/charts/issues/12371)
  # NB! It's highly recommended to change the default insecure rabbitmqErlangCookie value!
```

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Changelog
 
 ## In Development
-
+* Update `rabbitmq-ha` 3rd party chart from `1.44.1` to `1.46.1` (#158) (by @moonrail)
+* Disable newly introduced `rabbitmq-ha` prometheus operator by default (#158) (by @moonrail)
```
armab, line 5: This could be omitted as there's too verbose changelog for a single PR and we stick with the previous default.
```suggestion
```

armab (changes requested):
Thanks for PR! :+1: 

Just left a few minor comments to address.

values.yaml
```diff
@@ -458,7 +460,7 @@ rabbitmq-ha:
...
-  #rabbitmqErlangCookie: 8MrqQdCQ6AQ8U3MacSubHE5RqkSfvNaRHzvxuFcG
+  rabbitmqErlangCookie: 8MrqQdCQ6AQ8U3MacSubHE5RqkSfvNaRHzvxuFcG
```
armab, line 463: This was, in fact proposed in older PR and it's questionable default to me.

I realize that it could be useful for someone re-deploying (destroying/creating) rabbitmq-ha for many times in a row. And it also could take time to understand they need the same `rabbitmqErlangCookie` to make re-deployment not fail with the same PV/PVC.

But instead of forcing this rabbitmq cookie for everyone, we decided previously to hide the value under comment and include recommendation about why it could be useful or important. This way someone experiencing re-deployment issue would be able to consult with this Helm values hint.


Considering it's a second PR to enable `rabbitmqErlangCookie` by default, - let's do it :+1:

arms11: this was enabled in my cluster. I ran into this issue when we upgrade clusters (2 times) for Kubernetes version in EKS. AWS drains pods from old node groups and shift them to the new node groups. Even though all pods shift fine, due to the `rabbitmqErlangCookie` mismatch and not found in the helm chart, i was running into the issue which required me to delete the PVCs and then run `helm upgrade` to reconstruct the rabbitmq-ha cluster. Of course all that is a down-time. But since enabling that, i could see all pods shifted to the new nodes just fine and app was on-line the whole time! I would recommend it. Thanks!

values.yaml
moonrail: Suggested comment is added

armab:
Looks good!
Can you please also fix the `CHANGELOG.md` git conflict?

values.yaml
moonrail: Warning is added as well

CHANGELOG.md
moonrail: This line is now removed - wasn't sure about three lines of Changelog for one PR either when writing it

moonrail:
Rebased, conflict should be gone now

armab (approved):
:+1:

PR #159, opened 2020-11-20
Change pullPolicy to "IfNotPresent", as Docker-Hub Ratelimits now

moonrail:
Hello altogether

Docker-Hub introduced Ratelimiting: https://docs.docker.com/docker-hub/download-rate-limit/

While testing you will pull images a lot, also you may not want to always pull images, even on dev environments.

So I've figured, this would be a better default for most users.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -5,8 +5,8 @@
 ## Docker image settings, applied to all StackStorm pods
 ##
 image:
-  # Image pull policy. Change to "IfNotPresent" when switching to stable images
-  pullPolicy: Always
+  # Image pull policy
+  pullPolicy: IfNotPresent
   # st2 image repository. Set this to override the default ("stackstorm") or enterprise
   # docker image repository ("docker.stackstorm.com"). Applies to all st2 containers except
   # st2chatops and st2packs (which have their own override). This also does not impact 
@@ -99,7 +99,7 @@ st2:
       #repository: your-remote-docker-registry.io
       name: st2packs
       tag: latest
-      pullPolicy: Always
+      pullPolicy: IfNotPresent
       # Optional name of the imagePullSecret if your custom packs image is hosted by a private Docker registry behind the auth
       #pullSecret: st2packs-auth
 
@@ -403,7 +403,7 @@ st2chatops:
     #name: st2chatops
     ## Note that Helm templating is supported in this block!
     #tag: "{{ .Chart.AppVersion }}"
-    #pullPolicy: Always
+    #pullPolicy: IfNotPresent
   # Tested requested resource consumption for st2chatops & hubot in normal mode
   # Please adjust based on your conscious choice
   resources:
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -471,6 +472,9 @@ rabbitmq-ha:
   # Make sure to also change the rabbitmqMemoryHighWatermark following the formula:
   # rabbitmqMemoryHighWatermark = 0.4 * resources.limits.memory
   resources: {}
+  image:
+    # Image pull policy. Change to "IfNotPresent" when switching to stable images
+    pullPolicy: IfNotPresent
```
armab, line 477:
```suggestion
```
The pull policy is already `IfNotPresent` for the upstream chart. No reason to hardcode it here. Also the registry for rabbitmq-ha chart is outside of Docker Hub

values.yaml
```diff
@@ -489,6 +493,9 @@ etcd-operator:
   customResources:
     # create default etcd cluster
     createEtcdClusterCRD: true
+  image:
+    # Image pull policy. Change to "IfNotPresent" when switching to stable images
+    pullPolicy: IfNotPresent
```
armab, line 498:
```suggestion
```
The registry is quay.io for the etcd and `pullPolicy: IfNotPresent` in the upstream chart, no reason to include it here.

values.yaml
```diff
@@ -504,3 +511,6 @@ external-dns:
   aws:
     zoneType: "public"
   domainFilters: []
+  image:
+    # Image pull policy. Change to "IfNotPresent" when switching to stable images
+    pullPolicy: IfNotPresent
```
armab, line 516:
```suggestion
```

values.yaml
```diff
@@ -431,6 +431,7 @@ mongodb-ha:
   image:
     # StackStorm currently supports maximum MongoDB v4.0
     tag: "4.0"
+    pullPolicy: IfNotPresent
```
armab, line 434:
```suggestion
```

armab (changes requested):
Thanks for the contribution!

Looks good, just let's change the defaults for st2 images without touching the upstream chart settings as they already have `pullPolicy: IfNotPresent`.

We'll also need a respective CHANGELOG record for this PR.

moonrail:
@armab 
Makes perfectly sense to use upstream chart defaults - removed hardcoded pullPolicies for them.

values.yaml
```diff
@@ -6,7 +6,7 @@
 ##
 image:
   # Image pull policy. Change to "IfNotPresent" when switching to stable images
```
armab, line 8:
```suggestion
  # Image pull policy
```
Looks like this recommendation in comments could be removed now.

moonrail: Removed, also added Changelog

armab (approved):
Good stuff. Thanks a lot!

PR #161, opened 2020-11-20
Add post-delete to jobs hook definition to remove them on uninstall

### LINKED ISSUE #101: Some K8S resources do not get deleted after 'helm delete'.
Hello,

I would like to report an issue (I believe). After trying to delete the deployed helm chart, I can see that there are job pods remained. Here is an example of ```kubectl get po``` after ```helm delete --purge stackstorm``` on Helm 2 or ```helm delete stackstorm``` on Helm 3:
```
❯ kubectl get po   
NAME                                                              READY   STATUS                       RESTARTS   AGE
stackstorm-job-st2-apikey-load-j8rkv                              0/1     Completed                    0          6d21h
stackstorm-job-st2-key-load-wbdxp                                 0/1     Completed                    0          6d21h
stackstorm-job-st2-register-content-l6rmh                         0/1     Completed                    0          6d21h

```
Not only these pods, but also the jobs remain as well. on ```kubectl get jobs```, I can see:
```
❯ kubectl get job
NAME                                   COMPLETIONS   DURATION   AGE
stackstorm-job-st2-apikey-load         1/1           45s        6d21h
stackstorm-job-st2-key-load            1/1           9s         6d21h
stackstorm-job-st2-register-content    1/1           25s        6d21h
```

There are also PV,PVC objects remained but I believe that is desired?

Thanks!

moonrail:
Hello altogether

on Chart uninstalls all jobs remain present.

This Pull Request adds `post-delete` to `hook-delete-policy`, so helm deletes them as well.

Closes #101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## In Development
-
+* Add post-delete to jobs hook definition to remove them on uninstall (#161) (by @moonrail)
 
 ## v0.32.0
 * Fix a bug when datastore encrypted keys didn't work in scheduled rules. datastore_crypto_key is now shared with the ``st2scheduler`` pods (#148) (by @rahulshinde26)
--- a/templates/jobs.yaml
+++ b/templates/jobs.yaml
@@ -14,7 +14,7 @@ metadata:
     heritage: {{ .Release.Service }}
   annotations:
     helm.sh/hook: post-install, post-upgrade, post-rollback
-    helm.sh/hook-delete-policy: before-hook-creation
+    helm.sh/hook-delete-policy: before-hook-creation, post-delete
     helm.sh/hook-weight: "5"
 spec:
   template:
@@ -91,7 +91,7 @@ metadata:
     heritage: {{ .Release.Service }}
   annotations:
     helm.sh/hook: post-install, post-upgrade, post-rollback
-    helm.sh/hook-delete-policy: before-hook-creation
+    helm.sh/hook-delete-policy: before-hook-creation, post-delete
     helm.sh/hook-weight: "6"
 spec:
   template:
@@ -190,7 +190,7 @@ metadata:
     heritage: {{ .Release.Service }}
   annotations:
     helm.sh/hook: post-install, post-upgrade, post-rollback
-    helm.sh/hook-delete-policy: before-hook-creation
+    helm.sh/hook-delete-policy: before-hook-creation, post-delete
     helm.sh/hook-weight: "6"
 spec:
   template:
@@ -298,7 +298,7 @@ metadata:
     heritage: {{ .Release.Service }}
   annotations:
     helm.sh/hook: post-install, post-upgrade, post-rollback
-    helm.sh/hook-delete-policy: before-hook-creation
+    helm.sh/hook-delete-policy: before-hook-creation, post-delete
     helm.sh/hook-weight: "7"
 spec:
   template:


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## In Development
-
+* Add post-delete to jobs hook definition to remove them on uninstall
```
armab, line 4: Let's include the PR # to follow the same consistent format with the other records in the history.
```suggestion
* Add post-delete to jobs hook definition to remove them on uninstall (#161)
```

Also feel free to also include your github username as someone who contributed this fix, if you want.

moonrail: Ah, sure, I'll add it to the other PRs as well

armab:
What's the expected behavior for the `helm.sh/hook-delete-policy: post-delete`? Should it delete the abandoned jobs once the `helm delete <release_name>` ran?

So far I don't see that to happen in my minikube/helm environment.

moonrail:
Jipp, exactly.

On our kubernetes clusters all four jobs are left after `helm uninstall` without this hook.

We're using helm v3, so maybe there is a difference in behaviour to v2?

armab:
I just tested it in Helm 3 as well and don't see the jobs cleaned up with the `helm.sh/hook-delete-policy: before-hook-creation, post-delete`:

```
$ helm version
version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"clean", GoVersion:"go1.14.11"}

$ helm uninstall test
release "test" uninstalled

$ kubectl get all -l release=test
NAME                                      READY   STATUS      RESTARTS   AGE
pod/test-job-st2-apikey-load-n8457        0/1     Completed   0          6m52s
pod/test-job-st2-key-load-5svb2           0/1     Completed   0          5m20s
pod/test-job-st2-register-content-hhgcv   0/1     Completed   0          4m28s

NAME                                      COMPLETIONS   DURATION   AGE
job.batch/test-job-st2-apikey-load        1/1           91s        6m52s
job.batch/test-job-st2-key-load           1/1           51s        5m21s
job.batch/test-job-st2-register-content   1/1           68s        4m28s
```

Also can't find it documented for `helm.sh/hook-delete-policy` here: https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies
From the https://helm.sh/docs/topics/charts_hooks/#the-available-hooks `post-delete` is a hook name and could be used like this `"helm.sh/hook": post-delete`.

---

Can you please double check and test again in your environment? Any chance you did more modifications that are not included in the PR diff?

moonrail:
I thought, we had only this changed while testing - but you're right - according to helm's docs, this cannot have worked.
We'll test this again tomorrow from scratch.

moonrail:
@armab 
I've tested with three different installs - post-delete did never work, sorry for the inconveniences caused.
But also - jobs were never deleted (as well as PVs, but those I do like to remain :) ).

We could however add a `hook-succeeded` instead of it, so successful jobs would be cleaned up.
As these jobs should only run once after helm issued a hook this seems fine to me.

Or do you imagine users use these jobs via ad-hoc execution to e.g. revert issues with external (StackStorm Exchange, Git) packs, that had their actions modified via API? If thats even a use case.

armab:
Yeah, instead of deleting the successful job immediately, we want to keep it for the informational reasons so the user could potentially grab the logs and know which content was registered, packs, rbac rules applied, etc otherwise this Helm jobs magic remains unclear.

moonrail:
Alright, then this PR is not needed - closing it :)

armab:
Thanks for trying to help with that!

PR #162, opened 2020-11-23
Added option to define a pull secret for regular st2 images

moonrail:
Hello again,

as https://github.com/StackStorm/stackstorm-ha/pull/159 already was aimed at "mitigating" Docker Hubs new policies, this PR also shoots in this direction.

With this PR it is possible to define a registry pull secret for regular StackStorm images, in addition to a repository.

This helps if a user is using a private repository without anonymous access or if an user paid a subscription at e.g. Docker Hub and does not want to think about Ratelimiting, etc.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -12,6 +12,10 @@ image:
   # st2chatops and st2packs (which have their own override). This also does not impact 
   # dependencies such as mongo or redis, which have their own helm chart settings.
   repository: ""
+  # Image pull secret.
+  # May be required for public docker hub due to rate limiting or any private repository.
+  # See: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+  #pullSecret: "your-pull-secret"
 
 
 ##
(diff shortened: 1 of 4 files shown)


armab (approved):
That's a really well composed PR and helpful feature :+1: 

Thanks a lot!

values.yaml
```diff
@@ -12,6 +12,9 @@ image:
   # st2chatops and st2packs (which have their own override). This also does not impact 
   # dependencies such as mongo or redis, which have their own helm chart settings.
   repository: ""
+  # Image pull secret.
+  # May be required for public docker hub due to rate limiting or any private repository.
```
armab, line 16: I think having a link to instructions about how to set up the secret to pull images from private docker registry will help here: 

```suggestion
  # May be required for public docker hub due to rate limiting or any private repository.
  # See instructions: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
```

moonrail: Good idea - added it

PR #166, opened 2020-12-23
Allow multiple packs images

moonrail:
Hello Maintainers,

this PR adds the option to use multiple st2 packs-images.

This is a breaking change, as value `st2.packs.image` is replaced by `st2.packs.images`. Datatype changes from dict/hash to list of dicts/hashes.

This is useful, when a lot of different packs are being used, as a single container can result in multiple GB very quickly, mostly due to virtualenvs.

When using multiple st2 pack-images, the User can reduce possible downtimes due to reduced image size of new/changed containers. Also the build process can be a lot faster, if the User is providing custom packs.

Currently there is a fail message implemented in `templates/deployments.yaml` to prevent Users from accidential removal of their custom st2packs-image. It triggers, when `st2.packs.image` is defined and not empty. This should notify all Users about this change.

In #118 & #160 a centralization of packs-specific definitions for this chart was shown. As these PRs are still being discussed for other reasons, I've decided to include these centralizations in this PR.
This should make these PRs a little smaller and more compatible to each other, once they will or will not be finalized.

### DIFF
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -1,3 +1,8 @@
+# Notify users about breaking change regarding packs, to not destroy current installations
+{{- if and .Values.st2.packs.image }}
+{{- fail "Value st2.packs.image was renamed to st2.packs.images and is now a list of images" }}
+{{- end }}
+
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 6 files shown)


moonrail:
Rebased to current master, should now be mergeable again :)

armab (changes requested):
Thanks for PR! :+1:  Considering its breaking change nature looks good to mix with the other major behavior changes PRs.

Can you please sync the PR with the `master` one more time?

templates/deployments.yaml
```diff
@@ -1,3 +1,8 @@
+# Notify users about breaking change regarding packs, to not destroy current installations
+{{- if and .Values.st2.packs.image }}
+{{- fail "Value st2.packs.image was renamed to st2.packs.images and is now a list of images" }}
```
armab, line 3: Nice idea here! :+1:

moonrail:
Hey @armab 
Finally rebased and tested it again - should be mergable again :)

armab (approved):
Looks good!

Thanks a lot for contributing these changes! :+1:

PR #179, opened 2021-02-14
Added resource Requests values for St2 Pods

### LINKED ISSUE #5: Find resource Requests for all K8s pods/deployments
> Moved from https://github.com/StackStorm/k8s-st2/issues/25

Find resource requests and resource limits (at least Memory) for each container/pod in StackStorm cluster, allow modifying the values.

This is the best K8s practice which affects cluster capacity planning and so is must do for prod and all Deployment objects we write.

Capacity planning is based on defined K8s resources metadata and so it's important setting to configure.
Another quick example if an app introduces memory leak or rabbitmq app acquires all the memory it can see. If such pods deployed to the cluster with no limit set they can crash a node.

### Components
- [x] st2client
- [x] st2web
- [ ] st2auth
- [ ] st2api
- [ ] st2stream
- [ ] st2rulesengine
- [ ] st2timersengine
- [ ] st2workflowengine
- [ ] st2notifier
- [ ] st2sensorcontainer
- [ ] st2actionrunner
- [ ] st2garbagecollector

### A few references

https://github.com/StackStorm/stackstorm-enterprise-ha/blob/9e3caf754b14f1a20efe45f43bc3037f9f09c649/templates/jobs.yaml#L95-L96

https://github.com/StackStorm/stackstorm-enterprise-ha/blob/9e3caf754b14f1a20efe45f43bc3037f9f09c649/values.yaml#L210-L216

arms11:
Closes #5 
The values are added based on production monitoring average base on during peak/off-peak hours.
Please note: While I have not been able to test this in `minikube` due to unrelated reasons, below is the resource utilization of the `3` nodes cluster with `t3.large (2 CPUs/8GB Memory)` spec.

```❯ kubectl top node
NAME                             CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%
ip-192-168-137-84.ec2.internal   428m         22%    2587Mi          36%
ip-192-168-171-99.ec2.internal   279m         14%    3048Mi          42%
ip-192-168-99-86.ec2.internal    237m         12%    2640Mi          37%
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## In Development
+* Added resources requests cpu/memory values for St2 Pods (#179)
 * Implemented initContainers to wait for DB/MQ to be available for St2 Pods (#178)
 * Add option to define config.js for st2web (#165) (by @moonrail)
 
--- a/values.yaml
+++ b/values.yaml
@@ -119,8 +119,10 @@ st2:
         livenessProbe: {}
         readinessProbe: {}
         annotations: {}
-        # TODO: Find out recommended/default resources for this specific service (#5)
-        resources: {}
+        resources:
+          requests:
+            memory: "100Mi"
+            cpu: "50m"
         # Additional advanced settings to control pod/deployment placement
         affinity: {}
         nodeSelector: {}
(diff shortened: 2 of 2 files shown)


arms11:
@armab - Please review/guide. Thanks.

armab (commented):
Thanks for sharing the data here based on using the StackStorm-HA cluster :+1: 

Please also add a Changelog record and we should be good to go!

arms11:
Added Changelog entry and opened the pull request. Thanks.

armab (approved):
That worked well! :+1: 

Thanks for sharing the st2 pod usage data from the prod environment and codifying it here.
Tha'll help with proper resource allocation and scheduling in the wild.

PR #191, opened 2021-04-29
Add redis.staticID feature to allow constant sentinel ids 

manisha-tanwar:
Redis sentinel not able to elect master & all pods goes into CrashLoopBackOff when worker nodes are rebooted or 2 pods crashes for any reason.
[https://github.com/helm/charts/pull/19059](https://github.com/helm/charts/pull/19059)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,13 @@
 # Changelog
 
 ## In Development
-
+* Updated redis constant sentinel ID which will allow other sentinel peers to update to the new given IP in case of pod failure or worker node reboots. (#191) (by @manisha-tanwar)
 
 ## v0.60.0
 * Switch st2 version to `v3.5dev` as a new latest development version (#187)
 * Change st2packs definition to a list, to support multiple st2packs containers (#166) (by @moonrail)
 * Enabled RBAC/LDAP configuration for OSS version, removed enterprise flags (#182) (by @hnanchahal)
 * Fixed datastore_crypto_key secret name for rules engine (#188) (by @lordpengwin)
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -547,6 +547,7 @@ redis:
   # ## https://github.com/bitnami/charts/tree/master/bitnami/redis#master-slave-with-sentinel
   sentinel:
     enabled: true
+    staticID: true
```
armab, line 550: Would be nice to include a small code comment about this new value, perhaps with the original var description and maybe a bit with the context of how it helps in this situation.

armab (changes requested):
Thanks for the contribution!
The change looks good to me overall.

I provided one minor comment and we'll also need a Changelog.md update to merge this.

values.yaml
manisha-tanwar: Hi Does this looks fine now?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Change st2packs definition to a list, to support multiple st2packs containers (#166) (by @moonrail)
 * Enabled RBAC/LDAP configuration for OSS version, removed enterprise flags (#182) (by @hnanchahal)
 * Fixed datastore_crypto_key secret name for rules engine (#188) (by @lordpengwin)
+* Updated redis constant sentinel ID which will allow other sentinel peers to update to the new given IP in case of pod failure or worker node reboots. (#191) (by @manisha-tanwar)
```
manisha-tanwar, line 11: Added CHANGELOG too @armab 
Can you please review.

armab: Please move the Changelog record under the `In Development` section, otherwise looks good

armab (approved):
Looks good.

Thanks! :+1:

PR #182, opened 2021-02-16
Refactor Helm templates to configure LDAP/RBAC as OSS, remove Enterprise (v3.4)

hnanchahal:
Closes #134

PR #192, opened 2021-06-14
remove lingering reference to enterprise st2-license pullSecret

cognifloyd:
There was one more reference to `{{ .Release.name }}-st2-license` which was removed everywhere else in #182.
This removes the straggler.

As a follow-up to the already merged #182, this should not need a changelog entry. _edit: added changelog entry_


armab (changes requested):
Good find! :+1: 

@cognifloyd For better visibility, this still needs a Changelog record for history, to associate with the respective PRs #, and also because these two changes are part of the different releases.

Hope that makes sense.

cognifloyd:
I added a changelog entry.

armab (approved):
Thanks!

PR #193, opened 2021-06-14
Allow adding nodeSelector, affinity, tolerations to batch Jobs

cognifloyd:
Our cluster has dedicated nodes for StackStorm. So I need a way to specify the advanced placement settings (nodeSelector, affinity, tolerations) on the batch jobs as well.

- ~add advanced-pod-placement helper. This reduces duplication in the deployments file.~ _edit: moved to #216_
- add advanced pod/deployment placement settings for Jobs
- add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Add image.tag overrides for all deployments. (#200) (by @cognifloyd)
 * If your k8s cluster admin requires custom annotations (eg: to indicate mongo or rabbitmq usage), you can now add those to each set of pods. (#195) (by @cognifloyd)
 * Add optional hubot-scripts volume to st2chatops pod. To add this, define `st2chatops.hubotScriptsVolume`. (#207) (by @cognifloyd)
+* Add advanced pod placment (nodeSelector, affinity, tolerations) to specs for batch Jobs pods. (#193) (by @cognifloyd)
 
 ## v0.60.0
 * Switch st2 version to `v3.5dev` as a new latest development version (#187)
--- a/templates/jobs.yaml
+++ b/templates/jobs.yaml
@@ -77,6 +77,15 @@ spec:
           configMap:
             name: {{ .Release.Name }}-st2-rbac-mappings
       restartPolicy: OnFailure
+    {{- with .Values.jobs.nodeSelector }}
+      nodeSelector: {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.jobs.affinity }}
+      affinity: {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.jobs.tolerations }}
+      tolerations: {{- toYaml . | nindent 8 }}
+    {{- end }}
 
 {{- end }}
 ---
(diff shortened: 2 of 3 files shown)


cognifloyd:
pull-request-size bot says that this PR is Large because it deletes a lot of lines. In `templates/deployments.yaml` I added one line for every 12 lines I removed:

```diff
-    {{- with .Values.st2auth.nodeSelector }}
-      nodeSelector:
-{{ toYaml . | indent 8 }}
-    {{- end }}
-    {{- with .Values.st2auth.affinity }}
-      affinity:
-{{ toYaml . | indent 8 }}
-    {{- end }}
-    {{- with .Values.st2auth.tolerations }}
-      tolerations:
-{{ toYaml . | indent 8 }}
-    {{- end }}
+{{ include "advanced-pod-placement" .Values.st2auth | indent 6 }}
```

And then the template definition is just:
https://github.com/StackStorm/stackstorm-ha/blob/8356a5f75148836307e3ad84f3de1e8421d3dc1e/templates/_helpers.tpl#L166-L180

Then, I added 1 line per pod in the `templates/jobs.yaml`:
```diff
+{{ include "advanced-pod-placement" .Values.st2auth | indent 6 }}
```

Or to put that in context:
```yaml
---
apiVersion: ...
kind: ...
metadata:
  ...
spec:
  template:
    metadata:
      ...
    spec:
      ...
      containers:
        ...
      volumes:
        ...
{{ include "advanced-pod-placement" .Values.st2auth | indent 6 }}
```
results in:
```yaml
---
apiVersion: ...
kind: ...
metadata:
  ...
spec:
  template:
    metadata:
      ...
    spec:
      ...
      containers:
        ...
      volumes:
        ...
      nodeSelector:
        ...
      affinity:
        ...
      tolerations:
        ...
```

cognifloyd:
rebased on master

cognifloyd:
rebased on master

cognifloyd:
@armab how do you feel about using a helper for this block?

```diff
-    {{- with .Values.st2auth.nodeSelector }}
-      nodeSelector:
-{{ toYaml . | indent 8 }}
-    {{- end }}
-    {{- with .Values.st2auth.affinity }}
-      affinity:
-{{ toYaml . | indent 8 }}
-    {{- end }}
-    {{- with .Values.st2auth.tolerations }}
-      tolerations:
-{{ toYaml . | indent 8 }}
-    {{- end }}
+{{ include "advanced-pod-placement" .Values.st2auth | indent 6 }}
```


I feel this is a lot of repetition for each pod, so keeping `nodeSelector` `affinity` and `tolerations` inline does not add much value. If you don't want to use a helper, I can re-add all the repetition.

cognifloyd:
Here is what this PR would look like without the helper:
https://github.com/StackStorm/stackstorm-ha/compare/master...cognifloyd:advanced-pod-placement-repetitive

@armab which method do you prefer?

cognifloyd:
@armab I got this PR down to `size/M` please review. It no longer adds a helper and it does not reformat the deployments file.

armab (approved):
Looks nice! 
Thanks again for another enhancement :+1:

PR #194, opened 2021-06-14
minor quoting fix for MacVim highlighting

cognifloyd:
In MacVim, the highlighting is completely borked with double quotes inside other double quotes. So switch to single quotes which are equivalent.

There PR does not add, remove, or change any functionality. It just makes it a little nicer when reading or editing the templates.

This simple formatting change probably shouldn't need a changelog entry.


armab (approved):
Thanks!

Hope it'll be helpful for someone else!

PR #196, opened 2021-06-14
Add the ability to define imagePullSecrets for service account

cognifloyd:
If pods do not have imagePullSecrets, they get populated from the ServiceAccount[1]. In our k8s cluster, we prefer defining them on the ServiceAccount so that there are fewer places to update if a change is required.

Effectively, this means we can use `serviceAccount.pullSecret` instead of `image.pullSecret` in values.yaml.

[1] From [the k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#serviceaccount-admission-controller):
> 5. If the pod does not contain any imagePullSecrets, then imagePullSecrets of the ServiceAccount are added to the pod.

TODO:
- [x] PR description
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## In Development
 * Updated redis constant sentinel ID which will allow other sentinel peers to update to the new given IP in case of pod failure or worker node reboots. (#191) (by @manisha-tanwar)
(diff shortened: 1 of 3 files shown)


armab (approved):
Very helpful, thanks for the contribution! :+1:

PR #197, opened 2021-06-15
Allow setting an ST2_API_KEY for st2chatops instead of user/pass

cognifloyd:
Whenever I setup chatops, I prefer giving chatops an api_key instead of a username/password combo.

This PR allows us to add a previously generated st2 api_key as `st2chatops.api_key` in `values.yaml` (make sure to add the api_key object in `st2.apikeys` as well). Defining this api_key triggers replacing the `ST2_AUTH_USERNAME` and `ST2_AUTH_PASSWORD` environment vars with `ST2_API_KEY` but only for the st2chatops deployment.

### DIFF
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -1376,14 +1376,8 @@ spec:
       - name: st2chatops
         image: '{{ .Values.st2chatops.image.repository | default "stackstorm" }}/{{ .Values.st2chatops.image.name | default "st2chatops" }}:{{ tpl (.Values.st2chatops.image.tag | default .Chart.AppVersion) . }}'
         imagePullPolicy: {{ .Values.st2chatops.image.pullPolicy | default .Values.image.pullPolicy }}
+        {{- if not (hasKey .Values.st2chatops.env "ST2_API_KEY") }}
         env:
-        {{- if index .Values.st2chatops "api_key" }}
-        - name: ST2_API_KEY
-          valueFrom:
-            secretKeyRef:
-              name: {{ .Release.Name }}-st2-auth
-              key: chatops_api_key
-        {{- else }}
         - name: ST2_AUTH_USERNAME
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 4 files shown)


cognifloyd:
OK. I think I found a clean enough way to make st2chatops use an api_key. What do you think?

values.yaml
```diff
@@ -422,6 +422,9 @@ st2chatops:
   env:
     HUBOT_ADAPTER: slack
     HUBOT_SLACK_TOKEN: xoxb-CHANGE-ME-PLEASE
+  # If api_key is defined, then use this key instead of username/password for chatops to authenticate with StackStorm.
+  # Please make sure that this key gets added to st2.apikeys above so that chatops can use this key.
+  #api_key:
```
armab, line 427: Could the st2 API Key be configured via existing `env` definition?

```
env:
  HUBOT_ADAPTER: slack
  HUBOT_SLACK_TOKEN: xoxb-CHANGE-ME-PLEASE
  ST2_API_KEY: 123
```

cognifloyd: The problem is, I don't want ST2_AUTH_USERNAME or ST2_AUTH_PASSWORD exposed to the chatops process. So, adding the ST2_API_KEY var is only part of the issue.

templates/deployments.yaml
```diff
@@ -1377,6 +1377,13 @@ spec:
         image: '{{ .Values.st2chatops.image.repository | default "stackstorm" }}/{{ .Values.st2chatops.image.name | default "st2chatops" }}:{{ tpl (.Values.st2chatops.image.tag | default .Chart.AppVersion) . }}'
         imagePullPolicy: {{ .Values.st2chatops.image.pullPolicy | default .Values.image.pullPolicy }}
         env:
+        {{- if index .Values.st2chatops "api_key" }}
+        - name: ST2_API_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Release.Name }}-st2-auth
+              key: chatops_api_key
+        {{- else }}
```
cognifloyd, line 1386: I suppose we could do this instead:
```suggestion
        {{- if index .Values.st2chatops.env "ST2_API_KEY" }}
```
Would that be better?

values.yaml
cognifloyd: OK. I reworked this so ST2_API_KEY comes from `st2chatops.env` instead. That feels much cleaner. Thanks!

armab (approved):
Nice :+1: 
Thanks a lot.

BTW, I also remember that using ST2_API_KEY is more stable and recommended comparing to st2 username:password pair as st2chatops doesn't need to re-generate auth token (which sometimes fails on the javascript side).

Issue #132, opened 2020-05-13
appVersion, tag, and repository changes


ytjohn:
In #131 we added image.repository.  I'd also like to add an optional image.tag, while defaulting to Chart.AppVersion.   However, before I make the PR, I am seeing that st2chatops and st2sensorcontainer (which is really the user-provided custom st2packs container) both have their own "image.tag" and "image.repository" sections.

I see three paths to take:

1. Make the small change to Values.image.tag and leave st2chatops/st2sensorcontainer alone
2. Leave st2sensorcontainer alone and update st2chatops to default to top-level [image] section
2. Add standardized image.repository + image.tag to each st2* section in values.yaml, while defaulting to the top-level [images] section.

`image: "{{ template "imageRepository" . }}/st2actionrunner{{ template "enterpriseSuffix" . }}:{{ default .Chart.AppVersion .Values.image.tag }}"`

```
image:
  ...
  # custom tag for st2 docker images.
  # tag: "3.2dev-custom"
```

The end result would work like so:

```
$ helm template --set st2chatops.enabled=true . | grep 'image: .*/st2'
    image: "stackstorm/st2actionrunner:3.2dev"
        image: "stackstorm/st2auth:3.2dev"
        image: "stackstorm/st2auth:3.2dev"
        image: "stackstorm/st2api:3.2dev"
        image: "stackstorm/st2stream:3.2dev"
        image: "stackstorm/st2web:3.2dev"
        image: "stackstorm/st2rulesengine:3.2dev"
        image: "stackstorm/st2timersengine:3.2dev"
        image: "stackstorm/st2workflowengine:3.2dev"
        image: "stackstorm/st2scheduler:3.2dev"
        image: "stackstorm/st2notifier:3.2dev"
        image: "stackstorm/st2sensorcontainer:3.2dev"
        image: "stackstorm/st2actionrunner:3.2dev"
        image: "stackstorm/st2garbagecollector:3.2dev"
        image: "stackstorm/st2actionrunner:3.2dev"
        image: "stackstorm/st2actionrunner:3.2dev"
        image: "stackstorm/st2chatops:3.2dev"
        image: "stackstorm/st2actionrunner:3.2dev"
        image: "stackstorm/st2actionrunner:3.2dev"
        image: "stackstorm/st2actionrunner:3.2dev"
        image: "stackstorm/st2actionrunner:3.2dev"
        image: "stackstorm/st2actionrunner:3.2dev"
[...]

PR #200, opened 2021-06-18
Add custom image.tag for all pods

cognifloyd:
This adds an image.tag value for each of the pods.

Partially implements #132 (this only adds custom image.tag, not image.repository).

- add image.tag overrides for all deployments
- add image.tag overrides for all jobs


ericreeves:
+1.  Yes please.

This would allow me to easily utilize stable 3.4 containers with stackstorm-ha without having to maintain a patched Charts.yaml or a big Kustomize patch!

armab (approved):
Looks nice!

Thanks for the helpful addition :+1:

cognifloyd:
Oops. I missed one tag in a helper template. Pushed a follow up in #214.

PR #201, opened 2021-06-19
Add optional dnsPolicy, dnsConfig to all pods

cognifloyd:


cognifloyd:
rebased on master

cognifloyd:
@armab This is straight forward as well and should be quick to review.
It adds `dnsPolicy` and `dnsConfig` to all pods. It only adds lines; it does not change any existing lines.

Thanks again for all your reviews.

armab (approved):
Thanks!

Agree about making this setting common for all the Pods

PR #203, opened 2021-06-19
Refactor secrets and auto-generate password + ssh_key

cognifloyd:
Refactor secrets values to implement #14 and #16. Since these changes overlap, I combined them into one PR.

BREAKING CHANGE:
Move `secrets.st2.{username,password,ssh_key,datastore_crypto_key}` values into `st2.*` (implements #14).

Plus, auto-generate `password` and `ssh_key` secrets (implements #16), so that the stackstorm-ha installation is a bit more secure by default. On upgrade, use [`lookup`](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function) to pull the current secrets from k8s APIs.

Resolves: #14
Resolves: #16
Closes: #62


.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ jobs:
   helm-lint:
     working_directory: ~/stackstorm-ha
     docker:
-      - image: lachlanevenson/k8s-helm:v3.3.4
+      - image: lachlanevenson/k8s-helm:v3.4.2
```
cognifloyd, line 17: This change was triggered by: https://app.circleci.com/pipelines/github/StackStorm/stackstorm-ha/1192/workflows/9f1ccdab-bfef-4017-a98f-4b298bb49c75/jobs/4020

I'm using helm 3.5.4. e2e tests use the latest helm version, 3.6.0.

cognifloyd:
rebased on master

templates/secrets_st2auth.yaml
```diff
@@ -15,6 +22,10 @@ metadata:
...
-  username: {{ required "A valid secret 'st2.username' is required for StackStorm auth!" .Values.secrets.st2.username | b64enc | quote }}
-  # Password, used to login to StackStorm system (default: Ch@ngeMe)
-  password: {{ required "A valid secret 'st2.password' is required for StackStorm auth!" .Values.secrets.st2.password | b64enc | quote }}
+  username: {{ required "A valid secret 'st2.username' is required for StackStorm auth!" .Values.st2.username | b64enc | quote }}
+  # Password, used to login to StackStorm system (default: auto-generated)
+{{- if .Release.IsUpgrade }}
+  password: {{ index (lookup "v1" "Secret" .Release.Namespace $name).data "password" }}
+{{ else }}
+  password: {{ default (randAlphaNum 12) .Values.st2.password | b64enc | quote }}
+{{ end }}
```
armab, line 31: I hadn't chance to explore deeper the new `lookup` functionality.

How this thing will behave if user needs to override the previous password set?

Any edge cases?

templates/NOTES.txt
```diff
@@ -40,8 +40,8 @@ Ingress is enabled. You may access following endpoints:
 {{- end }}
 
 2. Login with the following credentials:
-username: {{ .Values.secrets.st2.username }}
-password: {{ .Values.secrets.st2.password }}
+username: {{ .Values.st2.username }}
+password: {{ .Values.st2.password }}
```
armab, line 44: Will it output the auto-generated password as well? if `Values.st2.password` is not initially set

cognifloyd: Good call. Hmm. This probably needs to do a `lookup` here.

templates/secrets_st2auth.yaml
cognifloyd: Hmm. We might need a flag like `resetPassword` which would make it skip the lookup. Something like this:
```
{{- if and .Release.IsUpgrade (not .Values.st2.resetPassword) }}
```

Wdyt?

cognifloyd: One edge case might be trying to update the password while upgrading from an existing install without setting that flag. I can't think of anything other edge cases. You?

cognifloyd: fixed

templates/NOTES.txt
cognifloyd: fixed

values.yaml
```diff
@@ -41,6 +41,25 @@ serviceAccount:
...
 ##
 st2:
+  # Username, used to login to StackStorm system
+  username: st2admin
+  # Password, used to login to StackStorm system (auto-generated by default)
+  # The password (set here or auto-generated) is preserved across upgrades.
+  #password: Ch@ngeMe
+  # To force the password to reset (using st2.password, or a newly generated one),
+  # you can use `helm upgrade --set st2.resetPassword=true`.
+  resetPassword: false
```
armab, line 51: I think the additional `resetPassword` functionality is a bit excessive and would be confusing.

The expected behavior is to always update the password or SSH key once the user changes it via the Helm values.

armab: Check the https://github.com/helm/helm/issues/3053#issuecomment-782837107 which tries to query the "previous" pass value instead of relying on `.Release.IsUpgrade`, it should help.

I think the expected and predictable behavior:
- No password/SSH key provided in Helm values = autogenerate it once, keep the state.
- If user changes Helm values unsetting the SSH key/password = just keep the previous state value, do nothing.
- If password/SSH key set in Helm values = always overwrite it with the new value.

WDYT? What are the other corner cases?

templates/secrets_st2auth.yaml
armab: Ah, replied in another thread: https://github.com/StackStorm/stackstorm-ha/pull/203/files#r660046629

values.yaml
cognifloyd: It sounds like you're thinking of something slightly different than the linked comment. Something like this, right?

```
{{- $previous := lookup "v1" "Secret" .Release.Namespace "foobar-secret" }}
data:
{{- if and $previous (not .Values.st2.password) }}
  password: {{ $previous.data.password }}
{{ else }}
  password: {{ default (randAlphaNum 12) .Values.st2.password | b64enc | quote }}
{{ end }}
```

And then similar logic for the ssh key.

Do we need to support rotating the auto-generated secrets? Or do we say that we only generate once, and any secret rotation must be managed via helm values?

armab: Yeah, I think rotation is the user's responsibility. It's already really good enough to have them auto-generated :)

cognifloyd: K. I removed `resetPassword` and updated the logic around password and ssh_key secrets.

armab:
Thanks! Looks great at a high level.

Based on the CI run (https://app.circleci.com/pipelines/github/StackStorm/stackstorm-ha/1251/workflows/dbc5fbc0-7557-4eb1-987e-a6d480366c3a/jobs/4191), looks like the st2 password is not showed up in the NOTES when doing the initial Helm chart install:


```
2. Login with the following credentials:
username: st2admin
password:
```

This is a high-profile feature that'll affect everyone and so a few more cycles of testing won't hurt, especially around the corner cases we mentioned.

cognifloyd:
Yeah. That's very sad - the lookup works on upgrade but not install.
Also, lookup always returns `nil` when using `helm template` to debug because helm is not supposed to contact the cluster during `template`.
I'll do a bit more research to see if there's a good way to grab the auto-generated value in multiple places. If you come across anything please let me know.

cognifloyd:
Helm templates all of the files at once, before applying the resources, so we cannot lookup in NOTES what was generated in another file. For that matter, during the same helm run, we can't lookup anything generated in one template file from another template file.

Also, NOTES is stored with the other release information. *By default* helm 3 uses secret storage for that, but also allows using configmap and postgresql storage backends. It makes sense to minimize the number of places that expose the plain secrets, without regard to how the cluster is [mis-]configured. So, I think using `kubectl` to retrieve the password instead of embedding it directly in the NOTES marginally improves security for the default st2 ha installation.

So, I changed the NOTES to explain how to retrieve the password with `kubectl`. Wdyt?

ericreeves:
I agree that it's a step forward to not just print the password in the NOTES.  Providing the kubectl command to retrieve it is far safer and nearly as simple from a user experience perspective. 

Printing passwords at every upgrade when they weren't requested is a bad thing.  They could get caught in log files and terminal output all over the place, and this is the admin password!

Great idea for a PR, and I think the implementation looks nice!

Any reason not to also implement the "auto-generate if not defined" behavior for the crypto key?

armab (approved):
Looks good :+1: Thanks @cognifloyd!

WIll run a few tests though before merging.
Would appreciate if @ericreeves you could help with more manual testing around possible edge cases.

PR #206, opened 2021-06-22
Allow adding postStartScript lifecycle hook to all deployments

### LINKED ISSUE #84: stanley_rsa is owned by root:root instead of stanley user in k8s
in the st2client pod sudo su - stanley then try to login somewhere using stanley_rsa key.  It won't work unless you are root or use sudo. 

```
whoami
stanley
stanley@stackstorm-st2client-b96dd9f76-t9kjp:~$ ls -l ~stanley/.ssh/
total 0
lrwxrwxrwx 1 root root 18 Sep  9 20:35 stanley_rsa -> ..data/stanley_rsa
stanley@stackstorm-st2client-b96dd9f76-t9kjp:~$ ssh undercloud.admin -i ~stanley/.ssh/stanley_rsa
The authenticity of host 'undercloud.admin' (10.75.163.57)' can't be established.
ECDSA key fingerprint is SHA256:Nr3mtGvtxhNNMeCzpy642VkASYji/u0xESuTTVKe4.
Are you sure you want to continue connecting (yes/no)? yes
Failed to add the host to the list of known hosts (/home/stanley/.ssh/known_hosts).
        ******** WARNING: UNAUTHORIZED PERSONS, DO NOT PROCEED ********
This system is intended to be used solely by authorized users in the course of
legitimate corporate business.  Users are monitored to the extent necessary to
properly administer the system, to identify the unauthorized users or users
operating beyone their proper authroity, and to investigate improper access or
use.  By accessing the system, you are consenting to this monitoring.
Additionally, users accessing this system agree that they understand and will
comply with all Verizon Information  Security and Privacy policies, including
policy statements, instructions, standards and guidelines.
        ******** WARNING: UNAUTHORIZED PERSONS, DO NOT PROCEED ********

Load key "/home/stanley/.ssh/stanley_rsa": Permission denied
stanley@undercloud.admin's password:
```

cognifloyd:
This builds on the postStart lifecycle hook from #219, but allows adding custom `postStartScript` (via values) to each deployment pod.

In my old 1ppc-based st2 cluster, the entrypoint script had some cluster-specific
customizations. With the current docker images, we don't use an entrypoint script.
Some of those customizations do not translate well to cacheable Dockerfile commands.

The `postStart` lifecycle event provides something similar in that I can add those
cluster-specific customizations when I can't bake the changes directly into custom
docker images.

For example, in the actionrunner pods, I have a script that:
  - configures ansible,
  - clones a git repo with ansible-related content,
  - configures ssmtp in the pod for sending email,
  - and runs a sanity check on some perl modules used by an internal pack
    (the perl modules themselves are baked into the docker image).
And in the chatops pod, our image has a slightly customized version of hubot that
needs an additional config file. That config file should not be baked into the image.

We also use the init script for rapid testing of changes in our development st2 cluster
before we bake them into the docker images (where baking the changes in makes sense).

Though these could potentially be done via initContainers, adding custom initContainers
becomes much more complex when you consider the required additional volumes and volumeMounts.

see:
https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/
https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Move st2-config-vol volume definition and list of st2-config-vol volumeMounts to helpers to reduce duplication (#198) (by @cognifloyd)
 * Fix permissions for /home/stanley/.ssh/stanley_rsa using the postStart lifecycle hook (#219) (by @cognifloyd)
 * Make system_user configurable when using custom st2actionrunner images that do not provide stanley (#220) (by @cognifloyd)
+* Allow providing scripts in values for use in lifecycle postStart hooks of all deployments. (#206) (by @cognifloyd)
 
 ## v0.60.0
 * Switch st2 version to `v3.5dev` as a new latest development version (#187)
--- a/templates/configmaps_post-start-script.yaml
+++ b/templates/configmaps_post-start-script.yaml
@@ -1,10 +1,12 @@
+{{- range tuple "st2auth" "st2api" "st2stream" "st2web" "st2rulesengine" "st2timersengine" "st2workflowengine" "st2scheduler" "st2notifier" "st2sensorcontainer" "st2actionrunner" "st2garbagecollector" "st2client" "st2chatops" -}}
+  {{- if or (index $.Values . "postStartScript") (eq . "st2actionrunner") (eq . "st2client") }}
 ---
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ $.Release.Name }}-st2actionrunner-post-start-script
+  name: {{ $.Release.Name }}-{{ . }}-post-start-script
   annotations:
-    description: Custom postStart lifecycle event handler script for st2actionrunner
+    description: Custom postStart lifecycle event handler script for {{ . }}
   labels:
     app: st2
     tier: backend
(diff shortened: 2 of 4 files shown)


cognifloyd:
re-pushed latest commit to restart e2e tests - its not working, so the postStartScript might be hanging st2actionrunner pod somehow.

cognifloyd:
I found the indentation issue that was hanging the e2e test. That's now fixed, so tests are passing.

I will push a changelog entry in the morning.
I'll also take a look at the tests to see about adding one for the ssh key file permissions.

cognifloyd:
This test seems to be passing now, so this PR demonstrably fixes #82 
https://github.com/StackStorm/stackstorm-ha/blob/913aed8c67767941af9dbbc9f5f33a1f0c138680/tests/st2tests.sh#L55-L65

So, this is now ready for review.

cognifloyd:
@armab I can also resolve #208 (make the "stanley" user configurable to accommodate custom st2actionrunner images) in this PR if you're ok with that. Otherwise I'll just add that in a follow-up PR once this one is merged.

cognifloyd:
Hmm. I should probably add a checksum annotation that covers the configmap template and the `postStartScript` value.

cognifloyd:
@armab I think this is the next one to review (please, when you have time).

It fixes the stanley_rsa file permissions issue using `lifecycle.postStart` and includes a test to make sure that the permissions stay fixed. It also allows for custom postStart scripts for the rest of the pods.

You might want to checkout the k8s docs in the PR description. :)

cognifloyd:
K. I split the stanley_rsa fix into #219. This is now only about adding the postStartScript to all the rest of the pods.

#206 builds on #219 now, so please review and merge #219 first.

armab:
I think it will solve one of the use cases that many requested, - installing packages for some of the st2 actions like, say, ansible instead of providing their own forked pre-baked image, both approaches with their pros/cons.

Oerall any customization before starting the st2 cluster makes sense and brings a lot of coniguration flexibility :+1:

cognifloyd:
@armab I'm glad you like the idea

armab (approved):
Nice addition for customizing the ST2 cluster environment.
Thanks for that!

PR #207, opened 2021-06-25
Add optional (advanced-only) hubot-scripts volume to st2chatops pod

cognifloyd:
This is for advanced use-cases only, and should generally not be used.
However, adding volumes is not possible from parent helm charts, so this is a limited, but useful, extension point.

I need to support an in-house acl solution for hubot. We use a hubot script to provide that feature. Given the current chart, there is no clean way to inject these scripts before chatops is started. This PR adds a simple, targeted way to enable this feature while keeping its footprint small.

NB. In my case, I have a custom docker image that includes the dependencies for my script, but there is also a config file that needs to be deployed via a volume of some kind, so embedding that in the docker image as well isn't feasible.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -513,7 +513,7 @@ st2chatops:
   # Advanced use-cases only. If defined, this hubot-scripts volume gets mounted to: /opt/stackstorm/chatops/scripts
   # This volume (using any k8s storage solution, including configmap) allows for hubot customization.
   # Most installations should not use this.
-  # Please refer to the hubot documentaiton for details on writing .js or .coffeescript hubot extensions.
+  # For details on writing .js or .coffeescript hubot extensions, see: https://hubot.github.com/docs/scripting/
   hubotScriptsVolume: {}
   # Additional advanced settings to control pod/deployment placement
   nodeSelector: {}
--- a/values.yaml
+++ b/values.yaml
@@ -510,6 +510,11 @@ st2chatops:
       memory: "50Mi"
       cpu: "5m"
   annotations: {}
+  # Advanced use-cases only. If defined, this hubot-scripts volume gets mounted to: /opt/stackstorm/chatops/scripts
+  # This volume (using any k8s storage solution, including configmap) allows for hubot customization.
+  # Most installations should not use this.
+  # Please refer to the hubot documentaiton for details on writing .js or .coffeescript hubot extensions.
+  hubotScriptsVolume: {}
   # Additional advanced settings to control pod/deployment placement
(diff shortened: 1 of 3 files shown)


cognifloyd:
rebased on master

values.yaml
```diff
@@ -498,6 +498,11 @@ st2chatops:
     requests:
       memory: "50Mi"
       cpu: "5m"
+  # Advanced use-cases only. If defined, this hubot-scripts volume gets mounted to: /opt/stackstorm/chatops/scripts
+  # This volume (using any k8s storage solution, including configmap) allows for hubot customization.
+  # Most installations should not use this.
+  # Please refer to the hubot documentaiton for details on writing .js or .coffeescript hubot extensions.
+  hubotScriptsVolume: {}
```
armab, line 505: I think it worth mentioning https://github.com/hubot-scripts here.
So users will have a better idea about the custom hubot scripts.

armab: How about overwriting the `external-scripts.json`?
That's the config you mentioned users would need to enable the external hubot scripts thy provided via volume.

cognifloyd: You only have to add things to `external-scripts.json` if they are *not* in the `scripts/` directory. So, we have to add `stackstorm-hubot` because it is in node_modules. When I drop my coffeescript file in `scripts/`, I do not register it in `external-scripts.json`.

I'll add the docs link soon.

cognifloyd: If you want to make `external-scripts.json` updatable, I would probably turn it into a configmap template so we can ensure hubot-stackstorm is registered. Would you like me to do that?

cognifloyd: OK. I made `external-scripts.json` updatable.

You still have to update the container though to install such scripts, so you could just edit `external-scripts.json` in the container itself. But I'm indifferent on external-scripts.json, because I don't need to touch it.

armab: > You only have to add things to external-scripts.json if they are not in the scripts/ directory. So, we have to add stackstorm-hubot because it is in node_modules. When I drop my coffeescript file in scripts/, I do not register it in external-scripts.json.

Makes sense, thanks for more details, I didn't know :+1: 

If what's included in the `scripts/` dir is already picked by the system, it's safer to revert that `external-scripts.json` customization as it has no play. Definitely would avoid shooting users in their foot.

Sorry for the confusion there.

cognifloyd: K, I dropped that change. This is ready now.

armab (approved):
:+1:

Issue #211, opened 2021-06-29
custom initContainer for register-content job


cognifloyd:
I need to disable most aliases and sensors that are enabled by default in system packs (`st2`, `linux`, and `packs`).
Since I need to adjust the system packs, I need this to do this whenever `helm install` or `helm upgrade` changes the system packs but before the content in the system packs gets registered.

Also, I want to ensure that these aliases are never enabled, so merely disabling them via API/UI once enabled is not good enough as that opens up a brief window where people can use `!pack install` and similar aliases that we don't want users to do via chatops.

I'm planning to use this with #199 (`st2.packs.volumes`), but it would also make sense when using `st2.packs.images`. With `st2.packs.images`, the custom images are copied into the `emptyDir` first, followed by the system packs (in the `packs-initContainers` helper template). So, custom images in `st2.packs.images` cannot change the system packs to disable unwanted aliases.

The helm chart uses `{{ .Release.Name }}-job-st2-register-content` to run `st2-register-content`. So, I think the ideal would be to allow for a custom `initContainer` on that `Job`. Then the question becomes: Where does such an initContainer definition go in values? `st2.packs.registerContentInitContainer`? `jobs.registerContent.initContainer`? Something else?

What should the UX for this look like?

cognifloyd:
OK. I added one possible approach in #213.

It would allow for something like this in values:
```yaml
jobs:
  registerContentJob:
    initCommand:
      - '/bin/bash'
      - '-c'
      - |
        sed -i -e '/^(\s*)enabled: true/$1enabled: false/' /opt/stackstorm/packs/packs/aliases/*.yaml
        sed -i -e '/^(\s*)enabled: true/$1enabled: false/' /opt/stackstorm/packs/linux/aliases/*.yaml
```

I'm not fond of `jobs.registerContentJob.initCommand` - is there somewhere better to put this?

cognifloyd:
Maybe `jobs.preRegisterContentCommand` would be a better name for this, or at least it is less nested.

PR #213, opened 2021-06-29
Add preRegisterContentCommand in an initContainer for register-content job

cognifloyd:
Allow a custom command to run just before `st2-register-content` that can disable aliases/sensors in system packs.

For example
```yaml
jobs:
  preRegisterContentCommand:
    - '/bin/bash'
    - '-c'
    - |
      sed -i -e 's/^\(\s*\)enabled: true/\1enabled: false/' /opt/stackstorm/packs/packs/aliases/*.yaml
      sed -i -e 's/^\(\s*\)enabled: true/\1enabled: false/' /opt/stackstorm/packs/linux/aliases/*.yaml
```

Closes: #211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 * Fix permissions for /home/stanley/.ssh/stanley_rsa using the postStart lifecycle hook (#219) (by @cognifloyd)
 * Make system_user configurable when using custom st2actionrunner images that do not provide stanley (#220) (by @cognifloyd)
 * Allow providing scripts in values for use in lifecycle postStart hooks of all deployments. (#206) (by @cognifloyd)
-* Add preRegisterContentCommand in an initContainer for register-content job (#213) (by @cognifloyd)
+* Add preRegisterContentCommand in an initContainer for register-content job to run last-minute content customizations (#213) (by @cognifloyd)
 
 ## v0.60.0
 * Switch st2 version to `v3.5dev` as a new latest development version (#187)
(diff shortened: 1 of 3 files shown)


cognifloyd:
@armab I just wrapped up this PR too. This is `size/S`, so hopefully it'll be quick to review.

cognifloyd:
@armab I think this is the easiest open k8s PR to merge.

Similar to the `postStart` change, this allows some init customization, but is restricted to only the register-content job.
Also, the `preRegisterContentCommand` runs in an `initContainer` because it has to run before the job itself, not in parallel (`postStart` runs in parallel with the entrypoint).

values.yaml
```diff
@@ -585,9 +585,21 @@ st2chatops:
...
   annotations: {}
   # Override default image settings (for now, only tag can be overridden)
+  # The Jobs use the st2actionrunner image
   image: {}
     ## Note that Helm templating is supported in this block!
     #tag: "{{ .Values.image.tag }}"
+  # If defined, this preRegisterContentCommand runs in an initContainer on the st2-register-content Job.
+  # The initContainer also uses the st2actionrunner image but runs this command instead of its entrypoint.
+  preRegisterContentCommand: []
+    # For example, to disable aliases in the "packs" system pack before content is registered:
```
armab, line 595: that's a good example :+1: 
Thanks for highlighting the use case here

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * Fix permissions for /home/stanley/.ssh/stanley_rsa using the postStart lifecycle hook (#219) (by @cognifloyd)
 * Make system_user configurable when using custom st2actionrunner images that do not provide stanley (#220) (by @cognifloyd)
 * Allow providing scripts in values for use in lifecycle postStart hooks of all deployments. (#206) (by @cognifloyd)
+* Add preRegisterContentCommand in an initContainer for register-content job (#213) (by @cognifloyd)
```
armab, line 20:
```suggestion
* Add preRegisterContentCommand in an initContainer for register-content job to run last-minute content customizations (#213) (by @cognifloyd)
```

armab (approved):
Thanks a lot for this enhancement!

cognifloyd:
Thank you!

PR #214, opened 2021-06-29
Fix tag in helper (follow up to #200)

cognifloyd:
I missed one tag in a helper template that needed the tag fix from #200.
This fixes that.

As a release hasn't been made yet, let the Changelog entry for #200 stand as is.

Follow up for #200


armab (approved):
Thanks for the follow-up!

Issue #205, opened 2021-06-22
Refactor spacing using nindent


cognifloyd:
It's difficult to read through our templates with many layers of yaml vs template indentation.
Also, it's hard to quickly see if the correct indentation will result due to `{{-` space chomping, which if too aggressive could accidentally put two separate lines on the same line.

The situation would be much better if we used [`nindent`](https://helm.sh/docs/chart_template_guide/function_list/#nindent) instead of [`indent`](https://helm.sh/docs/chart_template_guide/function_list/#indent) in various places so that we can keep our template indentation closer to the yaml indentation. `nindent` adds a newline at the start of the block, and then indents the block, which means you always know that the template will end up starting on the right line (not the previous line as when using `{{-`).
ie do this:

```diff
      imagePullSecrets:
      {{- if .Values.st2.packs.images }}
-{{- include "packs-pullSecrets" . | indent 6 }}
+       {{- include "packs-pullSecrets" . | nindent 6 }}
      {{- end }}
```

Such a change will invole a lot of whitespace change (which is the whole point), which would mean that any open PRs would need to be rebased to also use the improved template indentation.

armab:
Agreed :+1: 
It would be a nice code style enhancement indeed.

armab:
Implemented in https://github.com/StackStorm/stackstorm-ha/pull/215, thanks to @cognifloyd.

PR #198, opened 2021-06-16
Refactor: Use helper for st2-config-vol volume/volumeMounts definitions

cognifloyd:
This reduces duplication of the st2-config-vol definitions in `volumes:` and `volumeMounts:` blocks.

PR #215, opened 2021-07-02
yaml template style: use nindent to follow yaml indentation

cognifloyd:
Implements #205 for annotations, resources, and any other lines that are not covered in other PRs (#193, #198, #199)


armab (approved):
Thanks!

PR #218, opened 2021-07-06
Allow partitioning sensors with hash ranges across multiple deployments instead of one-sensor-per-pod

cognifloyd:
This allows us to specify `.Values.st2sensorcontainer.deployments` instead of hard-coding each sensor pod in `.Values.st2.packs.sensors`. It uses the ST2 hash range partitioning method to enable this behavior.
See: https://docs.stackstorm.com/reference/sensor_partitioning.html#hash

This may be helpful in private k8s clusters where hardware resources are constrained, so that only a few sensor containers can be running.

- sensor parition with hash ranges
- fix lint
- add changelog entry
- documentation

Props to @emptywee for figuring out how to calculate the hash ranges automatically. I translated what he did in a bash script into the helm chart.

TODO:

- [x] documentation in values.yaml and README (including adjustments to current docstrings)


cognifloyd:
K. I think this is ready for review. I renamed `st2sensorcontainer.replicas` to `st2sensorcontainer.deployments` to more accurately reflect how the hash range partitioning is done. Plus I added documentation in several places to explain the option.

cognifloyd:
Hmm. I think I can break this up into two PRs. I'll submit that in a few minutes.

cognifloyd:
#221 should be merged before this one will be ready to review.
#221 makes `st2sensorcontainer` provide defaults for each pod in `st2.packs.sensors` - this makes it easier to review as introducing `$sensor` and `$name` vars is easier to follow at a glance without introducing hash_ranges at the same time.

Then, this PR can focus on the sensor partitioning methods. It turns out, by adding hash ranges, there are 3 sensor modes (Yes, there *were* 2 modes before, it just wasn't obvious). Quoting from the docs I just updated in this PR:

> It is possible to run st2sensorcontainer(s) in one of these modes:
>   1) run all sensors in one pod (1 deployment with 1 pod, the default); or
>   2) run multiple sensors per pod (2+ deployments with 1 pod each) using hash range partitions; or
>   3) run one sensor per pod using `st2.packs.sensors`.
>
> To use the deployments (modes 1 and 2 in this list), `st2.packs.sensors` must be empty.
> For one-sensor-per-pod, define defaults here and add config for each sensor to `st2.packs.sensors`.

There was a default sensor container defined in `st2.packs.sensors` to ensure that there would be at least one sensor container.
Despite the inline documentation, that wasn't obvious - it just looked like an example that wasn't commented.

So, this improves the user experience by
1. using `st2sensorcontainer.deployments = 1` to cover that use case (all sensor in one pod), so that `st2.packs.sensors` is only for defining pods in the one-sensor-per-pod mode.
2. allow for scaling up sensors slightly by merely increasing `st2sensorcontainer.deployments` (hash range partitioning essentially becomes an implementation detail).
3. Then, if someone needs to go full HA with one sensor per pod, they can still do so with `st2.packs.sensors`.

cognifloyd:
I rebased this on master. This PR is much more focused than it was initially. This builds on #221, #222, #246 (all merged). It introduces a way to dynamically determine how to partition sensors with more than one sensor pod, without using one-sensor-per-pod mode.

cognifloyd:
@armab do you need anything else to be able to review this?

cognifloyd:
@armab RFR

cognifloyd:
I recommend reviewing these files first to see the documentation and user-facing bits first:
1. `CHANGELOG.md`
2. `README.md`
3. `values.yaml`

Then review this file backwards (review the changes near the end of the file and work your way to the beginning):
- `tempalates/deployments.yaml`

Reviewing `tempalates/deployments.yaml` from bottom to top will show you:
- how the sensorcontainer is configured (via command line or an initContainer-injected config file),
- the pod annotations: `stackstorm/sensor-mode` and `stackstorm/sensor-hash-ranges`,
- and finally, how the hash ranges are calculated and added to the `$sensor` map if needed.

ericreeves (approved):
Very nice.  I like that each sensorcontainer results in a separate deployment, which actually keeps operations tasks far simpler.

armab (commented):
Thanks for the PR!
I definitely like the idea that sensor partitioning is auto-detected just based on the number.

Though, it's definitely a complex feature: the implementation, configuration flow, adding even more moving pieces to an already complicated chart.
Historically, we didn't plan to implement something like this in the K8s for prod use, considering how K8s manages the Pods health, availability, self-healing, failover. And so one-sensor-for-container was considered as a production recommended way that's compatible with the K8s model.

Such a significant addition to the chart functionality needs a good testing and feedback.
It would be nice if someone from the community could try this PR and share their experience here, especially around the different sensor modes and how switching affects the environment.

cognifloyd:
> I definitely like the idea that sensor partitioning is auto-detected just based on the number.

Internally, we've had a script auto-calculating hash ranges on sensor pod startup since at least November 2017 with StackStorm 2.4.1 (I just looked through the git history). I hope this is useful for others as well, so I re-implemented that docker-entry-script using helm's template primitives.

> considering how K8s manages the Pods health, availability, self-healing, failover. And so one-sensor-for-container was considered as a production recommended way that's compatible with the K8s model.

In our environment, we provide the StackStorm platform and various team members can add packs that may or may not have sensors. They only need to learn the stackstorm layer, not the kubernetes layer as deploying StackStorm follows a completely different deployment model than everything else we deploy in kubernetes. So, defining the sensors is out-of-scope when installing StackStorm via the helm chart. All I can do is configure how many pods should be available for sensors, and other devs or devops people (who do not have access to deploy directly to kubernetes) will add their packs+sensors to StackStorm.

[hash ranges is the only method](https://docs.stackstorm.com/reference/sensor_partitioning.html) StackStorm supports for dynamic sensor partitioning. So, that's the method that I need to be able to install StackStorm in kubernetes.

As far as "how K8s manages the Pods health, availability, self-healing, failover", I chose to use deployments instead of replicas for this because it takes advantage of all of these kubernetes native features. The hash_range is node-specific configuration, so configuring all sensor nodes within the same deployment would have caused headaches if any particular sensor node crashes or needs to be moved to a different host for some reason (ie k8s follows a LIFO approach to replicas, you can't easily restart individual replicas). So, one-sensor-per-pod is not the only method that takes advantage of all of these k8s health-management features, hash-range based some-sensors-per-pod mode also maximizes use of the k8s-features given the constraint that I can't know how many sensors need to run at StackStorm installation time.

> Such a significant addition to the chart functionality needs a good testing and feedback.
> It would be nice if someone from the community could try this PR and share their experience here, especially around the different sensor modes and how switching affects the environment.

Configuring hash range partitioning has been a complex feature to configure on its own. This method simplifies that significantly when people are doing HA in k8s. So, maybe people will start preferring the hash-range strategy over all-sensors-in-one-pod mode once this is merged. Do you know anyone who is providing StackStorm as a platform-style service? Could you ping them to see if they can test this? Very few people seem to be working with stackstorm-ha, so I'm not sure who we can get to provide the feedback you want.

cognifloyd:
> Such a significant addition to the chart functionality needs a good testing

OK - I'm not sure how to add tests for this. Does anyone have any suggestions?

Is there a way to query StackStorm for which partitions are active? I can use `st2 sensor list` to list sensors, but I don't see a way to list the running sensorcontainers and which ranges they cover.

Maybe we can query kubernetes for the pods and check the annotations, but would such a test have any value? And how do I do kubernetes queries from within the bats tests? We don't have kubectl installed on the `st2actionrunner` image that the `st2tests-pod` uses... Would we have to manually craft k8s API requests to run via curl?

templates/deployments.yaml
```diff
@@ -1164,13 +1164,29 @@ spec:
       tolerations: {{- toYaml . | nindent 8 }}
     {{- end }}
 
+{{- $sensor_deployments_count := max 1 ($.Values.st2sensorcontainer.deployments|int) | int }}
 {{- $one_sensor_per_pod := not ($.Values.st2.packs.sensors | empty) }}
-{{- range ($one_sensor_per_pod | ternary ($.Values.st2.packs.sensors) (until 1)) }}
-  {{- $sensor := omit $.Values.st2sensorcontainer "name" "ref" "postStartScript" }}
+{{- $some_sensors_per_pod := gt $sensor_deployments_count 1 }}
+{{- if and $one_sensor_per_pod $some_sensors_per_pod }}
+  {{- fail "The sensor values are ambiguous. To use one-sensor-per-pod, use `st2.packs.sensors`. To use multiple-sensors-per-pod, use `st2sensorcontainer.deployments > 1`. Do not use both!" }}
```
cognifloyd, line 1171: I just added a safety check to help users avoid ambiguous sensor configuration. If `st2.packs.sensors` is not empty, then `st2sensorcontainer.deployments` must be the default (`1` or less).

cognifloyd (commented):
I just added unittests for this feature.

templates/deployments.yaml
armab: :+1:

README.md
```diff
@@ -123,6 +123,11 @@ which in our case is `Redis`.
 st2sensorcontainer manages StackStorm sensors: It starts, stops and restarts them as subprocesses.
 By default, deployment is configured with `1` replica containing all the sensors.
 
+You can increase the number of `st2sensorcontainer` pods by increasing the number of deployments.
+The replicas count is still only `1` per deployment, but the sensors are distributed between the deployments using
+[Sensor Hash Range Partitioning](https://docs.stackstorm.com/reference/sensor_partitioning.html#hash).
+The hash ranges are calculated automatically based on the number of deployments.
```
armab, line 129: The docs look really clear :+1: 
Hope more users will find it helpful to configure the sensor pods automatically.

armab (approved):
Thanks for all the efforts making it possible :100:

cognifloyd:
Woohoo! Thanks for the review!

PR #219, opened 2021-07-08
Fix stanley_rsa permissions via postStart pod lifecycle hook

### LINKED ISSUE #84: stanley_rsa is owned by root:root instead of stanley user in k8s

cognifloyd:
I extracted this change from #206.

Use the postStart lifecycle event in st2actionrunner and st2client pods to correct file permissions on the stanley ssh private key.
Includes a test to ensure the key permissions are correct.

see:
https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/
https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/

- Fix stanley_rsa file permissions with postStart lifecycle hook script for st2actionrunner and st2client
- bash needs to run the script directly not with -c
- make the ssh key writable so we can change permissions at runtime
- mount ssh-key to separate directory, cp, and fix permissions
- add test for stanley_rsa file ownership
- use st2 as test intermediary
- add changelog entry

Fixes #84

### DIFF
--- a/tests/st2tests.sh
+++ b/tests/st2tests.sh
@@ -30,6 +30,15 @@ load "${BATS_HELPERS_DIR}/bats-file/load.bash"
   assert_line --partial 'St2-Api-Key'
 }
 
+@test 'stanley_rsa file has correct permissions and ownership' {
+  local private_key="/home/stanley/.ssh/stanley_rsa"
+  assert_file_exist "${private_key}"
+  assert_file_not_empty "${private_key}"
+  assert_file_permission "500" "${private_key}"
+  assert_file_permission "400" "${private_key}"
+  assert_file_owner "stanley" "${private_key}"
+}
+
 @test 'st2 user can log in with auth credentials' {
   run st2 login ${ST2_AUTH_USERNAME} --password ${ST2_AUTH_PASSWORD} -w
   assert_success
--- a/tests/st2tests.sh
+++ b/tests/st2tests.sh
@@ -30,15 +30,6 @@ load "${BATS_HELPERS_DIR}/bats-file/load.bash"
   assert_line --partial 'St2-Api-Key'
 }
 
-@test 'stanley_rsa file has correct permissions and ownership' {
-  local private_key="/home/stanley/.ssh/stanley_rsa"
-  assert_file_exist "${private_key}"
-  assert_file_not_empty "${private_key}"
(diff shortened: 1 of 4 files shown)


tests/st2tests.sh
```diff
@@ -52,6 +52,18 @@ load "${BATS_HELPERS_DIR}/bats-file/load.bash"
   assert_line --partial 'succeeded: true'
 }
 
+@test 'stanley_rsa file has correct permissions and ownership' {
```
armab, line 55: :+1: Thanks for the tests!

armab (approved):
That's an elegant solution, thanks for contributing this fix :+1:

PR #220, opened 2021-07-08
Make system_user "stanley" configurable

### LINKED ISSUE #208: Allow replacing stanley with an alternate user
In my environment, the security team has strict requirements about the naming of service user accounts, so I can't us stanley.

In the `st2.config` value, I can change that user, and the ssh key as shown in https://github.com/StackStorm/st2/blob/master/conf/st2.conf.sample#L317-L321

```
[system_user]
# SSH private key for the system user.
ssh_key_file = /home/notstanley/.ssh/stanley_rsa
# Default system user.
user = notstanley
```

Then, I can use a custom `st2actionrunner` image that adds the `notstanley` account and home directory.
But, the name `stanley` is hard-coded in a couple of places. This is for the `st2actionrunner` pod, but the same applies to the `st2client` pod as well.
https://github.com/StackStorm/stackstorm-ha/blob/aabf3dcb201a86d5ddf6717cffc9dd76267181b5/templates/deployments.yaml#L1033-L1035
https://github.com/StackStorm/stackstorm-ha/blob/aabf3dcb201a86d5ddf6717cffc9dd76267181b5/templates/deployments.yaml#L1070-L1073

I think I would add new options in values:
```yaml
st2:
  # This mirrors the [system_user] section of st2.conf, but makes the values available for helm templating.
  # If you change the user, you must provide a customized st2actionrunner image that includes your user.
  system_user:
    user: stanley
    # if ssh_key_file is not defined, default is: /home/$system_user/.ssh/stanley_rsa
    #ssh_key_file:
```
Then, in st2-config configmap (in `st2.docker.conf`), add:
```
    {{- if index .Values "st2" "system_user" }}
    [system_user]
    user = {{ .Values.st2.system_user.user | default "stanley" }}
    {{- if index .Values "st2" "system_user" "ssh_key_file }}
    ssh_key_file = {{ .Values.st2.system_user.ssh_key_file }}
    {{- else }}
    ssh_key_file = /home/{{ .Values.st2.system_user }}/.ssh/stanley_rsa
    {{- end }}
    {{- end }}
```
With similar changes to `templates/deployments.yaml` for the ssh key mount paths.

I'm working on fixing the ssh key file permissions in #203 (to fix #84) so it would be easy to include this change in there, or I can wait till after that (or similar) is merged and add the configuration on top of it in a separate PR.

cognifloyd:
Currently the system_user is hard-coded to be "stanley". However, people can use a custom st2actionrunner image that does not have "stanley", and instead has some other user that StackStorm should use.

This PR allows us to inform helm about the alternate user, removing the hard-coded references to the "stanley" user.

- make system_user/stanley configurable
- changelog entry

Closes #208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Allow adding dnsPolicy and/or dnsConfig to all pods. (#201) (by @cognifloyd)
 * Move st2-config-vol volume definition and list of st2-config-vol volumeMounts to helpers to reduce duplication (#198) (by @cognifloyd)
 * Fix permissions for /home/stanley/.ssh/stanley_rsa using the postStart lifecycle hook (#219) (by @cognifloyd)
+* Make system_user configurable when using custom st2actionrunner images that do not provide stanley (#220) (by @cognifloyd)
 
 ## v0.60.0
 * Switch st2 version to `v3.5dev` as a new latest development version (#187)
--- a/templates/configmaps_post-start-script.yaml
+++ b/templates/configmaps_post-start-script.yaml
@@ -18,8 +18,12 @@ data:
   # see: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
   post-start.sh: |
     #!/bin/bash
-    mkdir -p /home/stanley/.ssh
-    cp -L /home/stanley/.ssh{-key-vol,}/stanley_rsa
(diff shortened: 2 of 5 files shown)


cognifloyd:
This builds on #219. I will remove this from "draft" mode once #219 is merged.

armab (approved):
Thanks for the multi-step contribution making the system user configuration possible :+1:

PR #221, opened 2021-07-17
Make st2.packs.sensors pull default values from st2sensorcontainer

cognifloyd:
This simplifies the one-sensor-per-pod method by pulling default values from `st2sensorcontainer`.

In other words, this makes `st2sensorcontainer` provide defaults for each pod in `st2.packs.sensors`.

### DIFF
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -982,13 +982,15 @@ spec:
     {{- end }}
 
 {{- range .Values.st2.packs.sensors }}
+  {{- $sensor := deepCopy $.Values.st2sensorcontainer | mustMergeOverwrite . }}
+  {{- $name := print "st2sensorcontainer" (include "hyphenPrefix" $sensor.name) }}
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ $.Release.Name }}-st2sensorcontainer{{ template "hyphenPrefix" .name }}
+  name: {{ $.Release.Name }}-{{ $name }}
   labels:
-    app: st2sensorcontainer{{ template "hyphenPrefix" .name }}
+    app: {{ $name }}
     tier: backend
     vendor: stackstorm
     chart: {{ $.Chart.Name }}-{{ $.Chart.Version }}
@@ -997,7 +999,7 @@ metadata:
 spec:
   selector:
     matchLabels:
-      app: st2sensorcontainer{{ template "hyphenPrefix" .name }}
+      app: {{ $name }}
       release: {{ $.Release.Name }}
   # https://docs.stackstorm.com/reference/ha.html#st2sensorcontainer
   # It is possible to run st2sensorcontainer in HA mode by running one process on each compute instance. Each sensor node needs to be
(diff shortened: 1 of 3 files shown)


cognifloyd:
@armab This is ready for review.

Aside - why didn't `pull-request-size` bot label this PR? I went ahead and added the `size/M` label myself.

templates/deployments.yaml
```diff
@@ -982,13 +982,18 @@ spec:
...
+  {{- $name := print "st2sensorcontainer" (include "hyphenPrefix" $sensor.name) }}
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ $.Release.Name }}-st2sensorcontainer{{ template "hyphenPrefix" .name }}
+  name: {{ $.Release.Name }}-{{ $name }}
   labels:
-    app: st2sensorcontainer{{ template "hyphenPrefix" .name }}
+    app: {{ $name }}
```
cognifloyd, line 996: This chunk was repeated on 5 lines, each of which has to be changed to use `$sensor.name`. So, I consolidated the name definition into one line (above).

templates/deployments.yaml
```diff
@@ -982,13 +982,18 @@ spec:
     {{- end }}
 
 {{- range .Values.st2.packs.sensors }}
+  {{- $sensor := omit $.Values.st2sensorcontainer "name" "ref" "postStartScript" }}
+  {{- range $key, $val := . }}
+    {{- $_ := set $sensor $key $val }}
+  {{- end }}
+  {{- $name := print "st2sensorcontainer" (include "hyphenPrefix" $sensor.name) }}
```
cognifloyd, line 989: The `$name` variable is only available (in-scope) inside the sensor `range` block.

templates/deployments.yaml
```diff
@@ -982,13 +982,18 @@ spec:
     {{- end }}
 
 {{- range .Values.st2.packs.sensors }}
+  {{- $sensor := omit $.Values.st2sensorcontainer "name" "ref" "postStartScript" }}
+  {{- range $key, $val := . }}
+    {{- $_ := set $sensor $key $val }}
+  {{- end }}
```
cognifloyd, line 988: This applies the default. I looked at using one of the `merge` functions to generate the `$sensor` dictionary, but all of the `merge` functions are deep merge functions which is undesirable. Using `set` like this allows us to replace keys instead of deep merging them.

values.yaml
```diff
@@ -482,8 +483,25 @@ st2actionrunner:
 
 # https://docs.stackstorm.com/reference/ha.html#st2sensorcontainer
 # Please see st2.packs.sensors for each sensor instance's config.
-# This contains settings that are common to all sensor pods.
+# This contains default settings for all sensor pods.
 st2sensorcontainer:
+  resources:
+    requests:
+      memory: "100Mi"
+      cpu: "50m"
```
cognifloyd, line 491: I just copied these defaults from what is currently in `st2.packs.sensors[0]`

cognifloyd (commented):
@armab here are a few comments explaining "why" on several parts of the PR.

armab (approved):
Thanks!

PR #239, opened 2021-09-06
imagePullSecrets is a top-level attribute of ServiceAccounts

### LINKED ISSUE #227: Fix indentation for ServiceAccount pullSecrets
I had the indentation wrong for `imagePullSecrets` in #196 

I have a branch prepared to fix it: https://github.com/cognifloyd/stackstorm-ha/tree/fix-sa-pullsecret
I will submit a PR later.

cognifloyd:
resolves #227


cognifloyd:
@armab We need this one for the next release.

armab (approved):
Looks good, but we'll still need a Changelog here :+1: 

Otherwise, it will be hard for people to fast-track the issues they experienced that might be silently fixed by us.

cognifloyd:
This fixes the syntax for #196 which has not been released yet. No one had been able to use this feature yet since it has been broken since it was introduced. Why do we need another entry?

armab:
There are folks who rely on the latest master chart and so the Changelog is always preferred as a way of tracking and sharing the progress and fixes.

PR #241, opened 2021-09-06
Automatically stringify st2.datastore_crypto_key and st2chatop.env values

### LINKED ISSUE #229: Cast values to string before b64encoding them in secrets
datastore_crypto_key and st2chatops secrets are both likely to have non-string content.
Examples:
- Using YAML or plain JSON formatting for datastore_crypto_key in values instead of a JSON-encoded string **wrapped in single quotes**.
- Using an integer for one of the chatops env vars (like `ST2_COMMANDS_RELOAD_INTERVAL: 10` instead of `ST2_COMMANDS_RELOAD_INTERVAL: "10"`)

We can easily cast these to string in the secret templates to fix this.

I have a branch prepared with the fix: https://github.com/cognifloyd/stackstorm-ha/tree/stringify
I'll submit a PR later.

cognifloyd:
datastore_crypto_key and st2chatops secrets are both likely to have non-string content.
Examples:

- Using YAML or plain JSON formatting for datastore_crypto_key in values instead of a JSON-encoded string **wrapped in single quotes**.
- Using an integer for one of the chatops env vars (like `ST2_COMMANDS_RELOAD_INTERVAL: 10` instead of `ST2_COMMANDS_RELOAD_INTERVAL: "10"`)

We can easily cast these to string in the secret templates to fix this.

Resolves #229


cognifloyd:
@armab I would like to include this one in the next release as well.

templates/secrets_datastore_crypto_key.yaml
```diff
@@ -20,6 +20,10 @@ metadata:
 type: Opaque
 data:
   # Datastore key used to encrypt/decrypt record for the KV store
+  {{- if kindIs "string" .Values.st2.datastore_crypto_key }}
   datastore_crypto_key: {{ .Values.st2.datastore_crypto_key | b64enc }}
+  {{- else }}
+  datastore_crypto_key: {{ .Values.st2.datastore_crypto_key | toJson | b64enc }}
```
armab, line 26: I wouldn't add an option for users to shoot themselves in their foot.
Datastore crypto key is an artifact that you generate and copy-paste, instead of trying to split it into pieces or re-format with yaml.

cognifloyd: Because I shot myself in the foot. JSON is valid YAML. If you don't surround the what you copy/paste with single quotes, it will be parsed into a dictionary and then things break.
This change makes the chart more forgiving.

armab: Does it also mean https://github.com/StackStorm/stackstorm-ha/blob/3dcd06b772ca5a2f4133a22fa9b382e03b5b422a/values.yaml#L50-L53 had initially wrong instructions?

cognifloyd: Yes. This change makes those instructions valid.

armab: Can we change the Helm values instructions? Similar to what we had for SSH key

```
datastore_crypto_key: |
  {"hmacKey": {"hmacKeyString": "", "size": 256}, "size": 256, "aesKeyString": "", "mode": "CBC"} 
```

templates/secrets_st2chatops.yaml
```diff
@@ -16,6 +16,6 @@ metadata:
 type: Opaque
 data:
 {{- range $env, $value := .Values.st2chatops.env }}
-  {{ $env }}: {{ $value | b64enc | quote }}
+  {{ $env }}: {{ $value | toString | b64enc | quote }}
```
armab, line 19: this looks good to me

templates/secrets_datastore_crypto_key.yaml
cognifloyd: Sure.
Which part of this change do you dislike? The `kindIs "String"` check, or passing the value through `toJson`?

cognifloyd: I updated the PR to drop `toJson` here and update the docs in values.yaml instead.

I would still like to understand your intuition - Why/how does serializing the `datastore_crypto_key` create "an option for users to shoot themselves in their foot"?

armab (approved):
Thanks a lot!

templates/secrets_datastore_crypto_key.yaml
armab: I think it comes from the expectations that the datastore crypto key is an immutable artifact that users just generate and copy-paste, instead of trying to split it into pieces, re-format, or modify. So it's best to keep that piece of data as is to avoid any further confusion or corner cases.

PR #242, opened 2021-09-06
Do not template rabbitmq secrets file unless rabbitmq subchart is enabled

### LINKED ISSUE #231: When RabbitMQ subchart is disabled, do not template its secrets file
We template a secrets file for use by the rabbitmq subchart.
I do not use that subchart, instead configuring st2 to use an externally managed rabbitmq cluster.
So, I have no need for that secrets file.

We can wrap that in a simple `if` block.
I have a branch prepared: https://github.com/cognifloyd/stackstorm-ha/tree/conditional-rabbitmq
I will submit a PR later.

cognifloyd:
We template a secrets file for use by the rabbitmq subchart.
I do not use that subchart, instead configuring st2 to use an externally managed rabbitmq cluster.
So, I have no need for that secrets file.

We can wrap that in a simple `if` block.

Resolves #231


cognifloyd:
@armab I would like to include this one in the next release as well.

armab (approved):
:+1:

PR #246, opened 2021-09-12
Improve one-sensor-per-pod UX: require "ref" in st2.packs.sensors

cognifloyd:
This makes it safer to configure `one-sensor-per-pod` mode, because it will error if `ref` is not defined when it should be. This way people cannot accidentally create multiple pods in `all-sensors-in-one-pod` mode.

`one-sensor-per-pod`:

-  Uses `st2.packs.sensors`, defaults come from `st2sensorcontainer`.
-  Require `ref` for every sensor in `st2.packs.sensors`.

`all-sensors-in-one-pod`:

-  Only use `st2sensorcontainer` for the only pod.
-  Ignore `st2.packs.sensors` in this mode.
-  No more magic first element in `st2.packs.sensors` to enable this.

Examples in `values.yaml` updated to be more focused.


armab (approved):
LGTM.

Thanks!

PR #248, opened 2021-09-12
Allow excluding default/sample RBAC files in values

### LINKED ISSUE #230: Do not include example RBAC files in values.
By default, we include `st2admin` and `stanley` users. However, those are optional and can be changed with custom images and modified values. If using an external auth source (I use ldap), those user accounts might not be allowed, or might be for something/someone else (thinking of `stanley`). So, some environments might not want to assign `stanley` to be a `system_user`.

For something like `st2.packs.images` it is easy to override the list with an empty list `[]`. That doesn't work for `st2.rbac.*` because `st2.rbac.roles` and `st2.rbac.assignments` are dicts/hashes which get merged together--`st2.rbac.mappings` is also a dict/hash, but it doesn't have any samples in it, so that needs no changes. So, there is not a clean way to remove `st2.rbac.roles["sample.yaml"]`, `st2.rbac.assignments["st2admin.yaml"]`, or `st2.rbac.assignments["stanley.yaml"]`.

It's fairly simple to drop those entries but leave them commented as examples in the values. That way there is no inadvertent granting of privileges to the wrong `stanley` user, or to a non-existent `st2admin` user.

I have a branch with this fix prepared: https://github.com/cognifloyd/stackstorm-ha/tree/no-default-rbac-files
I will submit a PR later.

cognifloyd:
As values dictionaries are merged, there is no way to remove unwanted rbac files like sample.yaml. And, we cannot assume that st2admin and stanley are present, because those are configurable.

So, we allow the user to set the sample/default files to an empty string to prevent them from inclusion in their stackstorm-ha cluster.

Resolves #230
Closes #247 (alternative implementation)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## In Development
 * Explicitly differentiate sensor modes: `all-sensors-in-one-pod` vs `one-sensor-per-pod`. Exposes the mode in new `stackstorm/sensor-mode` annotation. (#222) (by @cognifloyd)
 * Allow adding custom env variables to any Deployment or Job. (#120) (by @AngryDeveloper)
+* Set default/sample RBAC config files to "" (empty string) to prevent adding them. This is needed because they cannot be removed by overriding the roles/mappings values. (#247) (by @cognifloyd)
 
 ## v0.70.0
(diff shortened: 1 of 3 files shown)


armab (approved):
:+1: 

Thanks for the alternative implementation! This one provides much more flexibility.

PR #249, opened 2021-09-13
Template the contents of st2.config and the values in st2chatops.env.

### LINKED ISSUE #232: Feature: Use `tpl` to template config from values.
I do not use any of the external charts as that infrastructure is handled separately in our environments.
I also use `helm-secrets` to load secrets from our hashi `vault`.

For many of the values, grabbing the secret is very straightforward: `password: !vault secretv2/path/to/secret/password`.
But, for `st2.config` and for `st2chatops.env`, only parts of them need to be pulled from the vault (eg a redis password, rabbitmq credentials, or a slack token).

I would rather not store entire config files in `vault`. But, if we template these values with `tpl () .`, then I can easily reference my own values with something like this:
```
cognifloyd:
  rabbitmq:
    username: !vault secretv2/path/to/rmq/username
    password: !vault secretv2/path/to/rmq/password
  chatops:
    slack_token: !vault secretv2/path/to/slack/token
st2:
  config: |
    ...
    [messaging]
    url = amqp://{{ .Values.cognifloyd.rabbitmq.username }}:{{ .Values.cognifloyd.rabbitmq.password }}@my-rmq-host.example.com:5672/
    ...
st2chatops:
  env:
    ...
    HUBOT_SLACK_TOKEN: "{{ .Values.cognifloyd.chatops.slack_token }}"
    ...
```

I have a branch prepared to add this feature: https://github.com/cognifloyd/stackstorm-ha/tree/tpl-config
I will submit a PR later.

cognifloyd:
Resolves #232

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## In Development
 * Explicitly differentiate sensor modes: `all-sensors-in-one-pod` vs `one-sensor-per-pod`. Exposes the mode in new `stackstorm/sensor-mode` annotation. (#222) (by @cognifloyd)
 * Allow adding custom env variables to any Deployment or Job. (#120) (by @AngryDeveloper)
+* Template the contents of st2.config and the values in st2chatops.env. This allows adding secrets defined elsewhere in values. (#249) (by @cognifloyd)
 
 ## v0.70.0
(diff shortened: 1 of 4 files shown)


armab (approved):
Really helpful! :100:

PR #253, opened 2021-09-30
Use "--convert" when loading keys into datastore

### LINKED ISSUE #237: Feature: Allow passing raw YAML/JSON in values st2.keyvalue
Some keys that I need to load into the datastore (in `st2.keyvalue` in values) are easier to manage in YAML format.
StackStorm allows this by passing `--convert` or `-c` to `st2 key load` so that it casts the values to string before saving them in the datastore (all values in the datastore are strings).

I have a branch that adds this feature: https://github.com/cognifloyd/stackstorm-ha/tree/jobs_key_load_c
I will submit a PR later

cognifloyd:
Closes #237


armab (approved):
:+1:

PR #256, opened 2021-10-13
Add jobs.api_key so jobs can use api_key instead of username/password

### LINKED ISSUE #233: Feature: Allow hook/jobs to use an st2 api_key instead of user/pass

I will submit a PR later.

cognifloyd:
I always insist that automated jobs use API keys instead of username/password. This allows using an st2 API key with the helm hook jobs during `helm upgrade`. Just set `jobs.api_key` to a key that you created for helm to use.

Closes #233


rush-skills:
I am not sure if this fits, but I also have certain use cases (install packs and running other jobs using puppet) where using API-keys or auth tokens to authenticate is better than using username/password (since in my case the puppet install is not aware of the LDAP credentials of the service account, it is just injected in the config at a later stage). I have been playing around with the puppet module to add the ability to install packs with API-key instead of doing user auth (currently it requires authenticating as a Stanley system user to get the tokens, my environment has no Stanley user). So I won't say it's a single environment use case and might be something that is helpful for the community in general. 

Also again, it is okayish/safer to put an auth-token/API-key in the repo/conf/hiera than LDAP credentials of a user, at least in our case. 

For the chicken and egg problem - maybe we can make the scripts use username/password is api-key is missing, and use them to generate a "system" api-key that can be used for further transactions.

cognifloyd:
Closing in favor of #262

PR #261, opened 2021-10-26
allow bypassing generate-st2client-config init container for st2client and jobs pods

### LINKED ISSUE #233: Feature: Allow hook/jobs to use an st2 api_key instead of user/pass

cognifloyd:
This allows users who pass credentials via extra_volumes or envFromSecrets to bypass the `generate-st2client-config` initContainer.

This adds two boolean values: `st2client.generateST2ClientConfig` and `jobs.generateST2ClientConfig`. By setting `generateST2ClientConfig` to false, users can exclude the `generate-st2client-config` initContainer from st2client and/or jobs. Normally, that initContainer creates the `/root/.st2/config` file. Users may want to skip that initContainer when they:
- provide their own `/root/.st2/config` file via `extra_volumes`; or
- provide an `ST2_API_KEY` or `ST2_AUTH_TOKEN` via `envFromSecrets`.

Closes #233
Closes #256

Note: This PR is an alternate implementation of #256. (and I prefer this implementation over #256)
I want to provide the `ST2_API_KEY` via `envFromSecrets`, but the `/root/.st2/client` config with a `username` takes precedence over the `ST2_API_KEY` env var. When I looked into using `kustomize` I couldn't find a reliable way to remove the relevant initContainers. `jsonPatch` relies on array indexes which would be very brittle, and merge stragety patches can't remove items from an array. So, this PR provides a simple flag to exclude the initContainers in question.


armab (commented):
Thanks for the effort!

This looks better than the previous one with https://github.com/StackStorm/stackstorm-ha/pull/256, but still odd, focused on the specific rare use case.

What if we'll allow user to provide/template the `/root/.st2/config` file in the Helm values?
It can potentially bring more possibilities for customizations.
https://docs.stackstorm.com/reference/cli.html#configuration-file

Do you think it would be possible to put that together and make the specific case work for you?

cognifloyd:
Closing in favor of #262

PR #262, opened 2021-11-01
template st2client conf

### LINKED ISSUE #233: Feature: Allow hook/jobs to use an st2 api_key instead of user/pass

cognifloyd:
This moves the `/root/.st2/config` template (a bash heredoc) to values so that it can be adjusted based on install requirements. This also makes secrets in `envFromSecrets` available to the `generate-st2client-config` initContainers so that they can be used within the bash heredoc template. The template value is also passed through `tpl` to support helm templating for any non-secret config such as injecting the timezone.

Relevant st2 docs on the st2client config file: https://docs.stackstorm.com/reference/cli.html#configuration-file

Closes #233
Closes #256
Closes #261

Note: This PR is an alternate implementation of #256 and #261 based on @armab's feedback in #261.
I want to provide the `ST2_API_KEY` via `envFromSecrets`, but the `/root/.st2/client` config with a `username` takes precedence over the `ST2_API_KEY` env var. Templating the file allows me to either inject the api_key in the config file, or remove the credentials so that the `ST2_API_KEY` var is used.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -623,7 +623,7 @@ st2sensorcontainer:
 # The st2client deployment/pod simplifies ad-hoc administration.
 # st2client is a special purpose actionrunner pod, but you can customize it separately
 st2client:
-  # st2client config (~/.st2/config) template for jobs that need it.
+  # st2client config (~/.st2/config) template.
   # You can access env variables here because this is used in a bash heredoc.
   # For example, you could use a var injected with envFromSecrets.
   # Note that Helm templating is supported in this block!
--- a/values.yaml
+++ b/values.yaml
@@ -624,6 +624,7 @@ st2sensorcontainer:
 # st2client is a special purpose actionrunner pod, but you can customize it separately
 st2client:
   # st2client config (~/.st2/config) template.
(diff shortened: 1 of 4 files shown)


cognifloyd:
@armab RFR

values.yaml
```diff
@@ -623,6 +623,14 @@ st2sensorcontainer:
 # The st2client deployment/pod simplifies ad-hoc administration.
 # st2client is a special purpose actionrunner pod, but you can customize it separately
 st2client:
+  # st2client config (~/.st2/config) template.
+  # You can access env variables here because this is used in a bash heredoc.
```
armab, line 627: Nice!
Let's just include a link to https://docs.stackstorm.com/reference/cli.html#configuration-file in comments for both Helm value definitions.

That'll help users find the use cases and configuration examples easily.

armab (approved):
Really nice feature now that would work for everyone!
Thanks a lot for finding the way here :+1: 

Left one minor comment to address, otherwise looks great.

values.yaml
cognifloyd: done

Issue #236, opened 2021-08-05
Feature: Allow serving https from the st2web pod


cognifloyd:
Currently we have the st2web pods serving HTTP.
In our environment, we have another nginx instance that is the primary entry point for the domain associated with st2. This nginx host is outside the k8s cluster and then forwards the traffic to the st2web pods in the cluster.
I want the traffic going into the cluster to also be HTTPS, though I'm fine with a generated certificate.

I have a branch prepared to add this feature: https://github.com/cognifloyd/stackstorm-ha/tree/st2web-https
My branch allows people to enable HTTPS in st2web, and if an `ssl_certificate` + `ssl_certificate_key` are not provided in values, then a new CA  and a signed cert (signed by the new CA) are generated in the helm templates.

I will submit a PR later.

armab:
We had HTTPS for st2web before, but after several discussions in the past moved away from this model.

The current way is to keep the st2web HTTP only, allowing users to rely on ingress controllers, load-balancers, and everything else to configure HTTPS on top of it.

Keeping the secure layer outside of the stackstorm-ha scope provides the separation of concerns and minimizes the amount of moving pieces and configuration. In K8s considering plenty of great solutions, it's something that's best configured and controlled outside of st2web.

cognifloyd:
I went through all the git history and the related issues+PRs to understand the change to HTTP, and the conversation around it, before I worked on adding HTTPS support back in. Hopefully I didn't miss something.

I think there's a clean way to support both that offers a simple interface for chart users.

PR #264, opened 2021-11-11
Allow st2web to serve HTTPS

cognifloyd:
The Docker image allows serving HTTPS when environment var `ST2WEB_HTTPS=1`. Setting `ST2WEB_HTTPS: "1"` in `st2web.env`, however, breaks the `st2web` k8s deployment & service because the `livenessProbe` and the service are still trying to access HTTP on port 80.

To make `ST2WEB_HTTPS=1` work correctly, 2 things are required:
1. change the protocol and port if `ST2WEB_HTTPS=1`.
2. add `extra_volumes` (like several other deployments) so users can mount ssl certs that they provide. Managing any kubernetes secrets for those certs is out-of-scope for this chart, similar to anything else provided via `extra_volumes`.

This PR makes those changes, but retains the current HTTP-by-default behavior.

Closes: #236


ericreeves (approved):
Elegantly introduced feature, and backward-compatible.  This is a solid feature for those running in stacks who can't just shovel the work off to a cloud load balancer.

values.yaml
```diff
@@ -295,8 +295,15 @@ st2web:
   affinity: {}
   env: {}
   # HTTP_PROXY: http://proxy:1234
+  ## Advanced: To configure st2web to serve HTTPS (port 443) instead of HTTP (port 80)
+  ## provide ssl certs via extra_volumes, and add the ST2WEB_HTTPS env var here:
+  # ST2WEB_HTTPS: 1
```
armab, line 300: It makes sense to quickly mention in the comments about the alternatives like LoadBalancer or Ingress Controller, which would be a more K8s-native approach to this problem.

armab (approved):
I thought about discussing more the approach in this PR as we've removed it before in favor of LoadBalancer or Ingress Controller which is more K8s native and flexible.

Because we have this functionality still present in the Docker images and enabling it really easy, - let's do it! 

Thanks for the PR :+1:

values.yaml
cognifloyd: done

PR #265, opened 2021-11-11
Enable extra post-install and post-upgrade helm-hook jobs for environment-specific config

### LINKED ISSUE #234: Feature: Allow users to define custom helm hook jobs to run on install/upgrade/etc
Related to #233, I need to be able to generate an api_key on install before any of the other helm hook jobs run.
I also need to run a workflow to save the enabled/disabled state of aliases before an upgrade, and restore them after an upgrade.
Ideally, helm should kick off those tasks.

Helm hooks can be applied to many k8s objects, but I only need to run Jobs, so I think we should offer a very narrow way to define hook jobs without going to the trouble of creating a parent chart.

I have a branch prepared to add this `jobs.extra_hooks` feature: https://github.com/cognifloyd/stackstorm-ha/tree/jobs_extra_hooks
I will submit a PR later.

cognifloyd:
Resolves #234

There is a lot of boilerplate for custom StackStorm hook-jobs. That boilerplate is likely to become out-of-date as we refine the chart, so managing that boilerplate within this chart alongside our other jobs would be ideal.

Since there is so much chart-specific boilerplate, parent charts would have to depend on a lot of implementation details making them very brittle. Parent charts are much better suited to jobs that don't need access to most of the packs, configs, configmaps, and secrets that this chart provides.

These hook-jobs can be used for st2 installation-specific jobs like:
- running a pre-upgrade st2 workflow that notifies on various channels that the upgrade is happening,
- running post-upgrade smoke tests to ensure st2 can connect to vital services (vault, kubernetes, aws, etc),
- running a pre-upgrade st2 workflow that pauses long-running workflows,
- running a post-upgrade st2 workflow that resumes long-running workflows,
- running one-time post-install configuration (such as generating dynamic secrets in the st2kv datastore),
- etc.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -902,6 +902,7 @@ jobs:
   # See available hooks list: https://helm.sh/docs/topics/charts_hooks/#the-available-hooks
   extra_hooks: []
     # Each item in extra_hooks must define a 'name' and the "helm.sh/hook" value in 'hook'.
+    # Note that Helm templating is supported in the 'command' block!
     #
     # - name: init-workflow  # required name
     #   hook: post-install   # required "helm.sh/hook"
--- a/values.yaml
+++ b/values.yaml
@@ -895,7 +895,32 @@ jobs:
   # For example, if an upgrade only touches RBAC config, use this to disable other jobs:
   #   helm upgrade ... --set 'jobs.skip={apikey_load,key_load,register_content}'
   skip: []
-
+  #
+  # Advanced: Add extra Helm hook Jobs
+  # These hook jobs will use the same settings (eg image, annotations, pod placement) as the other jobs.
+  # They will have st2 cli configured, st2.conf files, and packs volumes mounted.
+  # See available hooks list: https://helm.sh/docs/topics/charts_hooks/#the-available-hooks
+  extra_hooks: []
+    # Each item in extra_hooks must define a 'name' and the "helm.sh/hook" value in 'hook'.
+    #
+    # - name: init-workflow  # required name
+    #   hook: post-install   # required "helm.sh/hook"
+    #   hook_weight: 10      # optional hook_weight (defaults to 10)
+    #   resources: {}        # optional definition of resources to request
+    #   command:             # required command to run
+    #     - st2
+    #     - run
+    #     - --tail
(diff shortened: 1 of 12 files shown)


ericreeves (approved):
This is a great PR that can resolve a lot of automation use cases.

Very clear implementation.  

Thanks!

cognifloyd:
Rebased on master

cognifloyd:
Well that was nice. Adding unit tests caught a few issues with this PR. I think we need unit tests for env and volumes before merging this. Once I've prepared those, I'll push a PR.

cognifloyd:
rebased and squashed into a logical set of commits.

cognifloyd:
Rebased and added `extra_hooks` jobs to existing tests to ensure this is working correctly. That uncovered a few issues that are now resolved.

values.yaml
```diff
@@ -895,7 +895,33 @@ jobs:
   # For example, if an upgrade only touches RBAC config, use this to disable other jobs:
   #   helm upgrade ... --set 'jobs.skip={apikey_load,key_load,register_content}'
   skip: []
-
+  #
+  # Advanced: Add extra Helm hook Jobs
+  # These hook jobs will use the same settings (eg image, annotations, pod placement) as the other jobs.
+  # They will have st2 cli configured, st2.conf files, and packs volumes mounted.
+  # See available hooks list: https://helm.sh/docs/topics/charts_hooks/#the-available-hooks
```
armab, line 902: This is powerful, but I think the use case examples would be nice to include somewhere. Otherwise, very few would know about it and use it.
Perhaps a Readme section describing this feature with some examples?

armab (approved):
:+1: 

I think it is worth a release after merging this PR and looking at how many features already accumulated.

values.yaml
cognifloyd: That makes sense. I will add a README section.

cognifloyd:
K. README updated in https://github.com/StackStorm/stackstorm-ha/pull/265/commits/e4b72dc2162824a64ec8150d4e8362c39e645e29

cognifloyd:
I'll merge now. We can revise the README in a follow-up if needed.

Issue #225, opened 2021-07-28
Feature: Auto-generate datastore_crypto_key if not provided


cognifloyd:
It would be nice to auto-generate the `datastore_crypto_key` as part of the helm install.
All of the templates get templated at once, so using the `st2-generate-symmetric-crypto-key` utility (as described in [the datastore doc](https://docs.stackstorm.com/datastore.html#securing-secrets-admin-only)) cannot be encapsulated within the chart. Instead, we need to be able to generate the key using [helm](https://helm.sh/docs/chart_template_guide/function_list/) or [sprig](https://masterminds.github.io/sprig/) functions.

Internally, `st2-generate-symmetric-crypto-key` uses `st2common.util.crypto.AESKey.generate(key_size=256)`. That returns a json document with two keys that need to be generated: `hmacKey.hmacKeyString` and `aesKeyString`.

Those strings are _"Base64 web safe encoding [with] suppress[ed] padding characters (=)."_ Here "web safe encoding" means _"Uses URL-safe alphabet: - replaces +, _ replaces /."_ (from the docstrings)

Essentially, those two strings are generated with this snippet of python:
```
base64.urlsafe_b64encode(os.urandom(int(256 / 8))).decode("utf-8").replace("=", "")
```

Looking over the helm and sprig docs, I believe we can generate those strings with this (32 = 256/8):
```
"{{ randBytes 32 | replace '+' '-' | replace '_' '/' | replace '=' '' }}"
```

So, in values, the `datastore_crypto_key` could be generated with something like this:
```yaml
st2:
  datastore_crypto_key:
    hmacKey:
      # 32 bytes = 256 bits / 8 bits/byte
      # this formula is based on an st2-specific version of python's base64.urlsafe_b64encode.
      # randBytes returns a base64 encoded string.
      hmacKeyString: "{{ randBytes 32 | replace '+' '-' | replace '_' '/' | replace '=' '' }}"
      size: 256
    aesKeyString: "{{ randBytes 32 | replace '+' '-' | replace '_' '/' | replace '=' '' }}"
    mode: CBC
    size: 256
```

When this gets integrated in [`templates/secrets_datastore_crypto_key.yaml`](https://github.com/StackStorm/stackstorm-ha/blob/master/templates/secrets_datastore_crypto_key.yaml), we also need to account for converting from the dict/hash to a json-encoded string. So, maybe something like:
```yaml
[...]

PR #266, opened 2021-11-11
Generate datastore_crypto_key on install if not provided

cognifloyd:
OK. I wasn't going to implement this, but I did it anyway.

Other installation methods can call `st2-generate-symmetric-crypto-key` on install
to generate the key. But we need to put it in a kubernetes secret, so that is too
late to use for a new installation. Instead, we can generate the crypto key with
helm primitives.

Closes: #225

### DIFF
--- a/conf/datastore_crypto_key.yaml
+++ b/conf/datastore_crypto_key.yaml
@@ -1,0 +1,13 @@
+# This is used to generate st2.datastore_crypto_key on install if not defined in values.
+
+# The formula is based on an st2-specific version of python's base64.urlsafe_b64encode
+# randBytes returns a base64 encoded string
+# 32 bytes = 256 bits / 8 bits/byte
+
+aesKeyString: '{{ randBytes 32 | replace "+" "-" | replace "_" "/" | replace "=" "" }}'
+mode: CBC
+size: 256
+
+hmacKey:
+  hmacKeyString: '{{ randBytes 32 | replace "+" "-" | replace "_" "/" | replace "=" "" }}'
+  size: 256
--- a/templates/secrets_datastore_crypto_key.yaml
+++ b/templates/secrets_datastore_crypto_key.yaml
@@ -24,7 +24,7 @@ data:
 {{- $previous := lookup "v1" "Secret" .Release.Namespace $name }}
 {{- if and $previous (not .Values.st2.datastore_crypto_key) }}
   datastore_crypto_key: {{ $previous.data.datastore_crypto_key }}
-{{- elif .Values.st2.datastore_crypto_key }}
+{{- else if .Values.st2.datastore_crypto_key }}
   datastore_crypto_key: {{ .Values.st2.datastore_crypto_key | b64enc }}
 {{- else }}
   datastore_crypto_key: {{ tpl (.Values.default_datastore_crypto_key | toJson) . | b64enc }}
--- a/templates/secrets_datastore_crypto_key.yaml
+++ b/templates/secrets_datastore_crypto_key.yaml
@@ -20,6 +20,10 @@ metadata:
 type: Opaque
 data:
   # Datastore key used to encrypt/decrypt record for the KV store
+{{- if kindOf .Values.st2.datastore_crypto_key | eq "string" }}
   datastore_crypto_key: {{ .Values.st2.datastore_crypto_key | b64enc }}
+{{- else }}
+  datastore_crypto_key: {{ tpl (.Values.st2.datastore_crypto_key | toJson) . | b64enc }}
+{{- end }}
 
 {{- end }}
--- a/templates/secrets_datastore_crypto_key.yaml
+++ b/templates/secrets_datastore_crypto_key.yaml
@@ -27,6 +27,6 @@ data:
 {{- else if .Values.st2.datastore_crypto_key }}
   datastore_crypto_key: {{ .Values.st2.datastore_crypto_key | b64enc }}
 {{- else }}
-  datastore_crypto_key: {{ tpl (.Values.default_datastore_crypto_key | toJson) . | b64enc }}
+  datastore_crypto_key: {{ tpl (.Values.default_datastore_crypto_key | toRawJson) . | b64enc }}
 {{- end }}
 
--- a/templates/secrets_datastore_crypto_key.yaml
+++ b/templates/secrets_datastore_crypto_key.yaml
@@ -22,10 +22,10 @@ type: Opaque
 data:
   # Datastore key used to encrypt/decrypt record for the KV store
 {{- $previous := lookup "v1" "Secret" .Release.Namespace $name }}
-{{- if and $previous (not .Values.st2.datastore_crypto_key) }}
-  datastore_crypto_key: {{ $previous.data.datastore_crypto_key }}
-{{- else if .Values.st2.datastore_crypto_key }}
+{{- if .Values.st2.datastore_crypto_key }}
   datastore_crypto_key: {{ .Values.st2.datastore_crypto_key | b64enc }}
+{{- else if $previous }}
+  datastore_crypto_key: {{ $previous.data.datastore_crypto_key }}
 {{- else }}
   datastore_crypto_key: {{ tpl (.Files.Get "conf/datastore_crypto_key.yaml") . | fromYaml | toRawJson | b64enc }}
 {{- end }}
--- a/templates/secrets_datastore_crypto_key.yaml
+++ b/templates/secrets_datastore_crypto_key.yaml
@@ -2,7 +2,7 @@
 {{- $deprecated_crypto_key := (default (dict) (default (dict) .Values.secrets).st2).datastore_crypto_key }}
 {{- if $deprecated_crypto_key }}
 {{- fail "Please update your values! The datastore_crypto_key value moved from secrets.st2.* to st2.*" }}
-{{- else if ne "disable" .Values.st2.datastore_crypto_key }}
+{{- else if ne "disable" (default "" .Values.st2.datastore_crypto_key) }}
 ---
 apiVersion: v1
 kind: Secret
--- a/templates/secrets_datastore_crypto_key.yaml
+++ b/templates/secrets_datastore_crypto_key.yaml
@@ -27,6 +27,7 @@ data:
 {{- else if .Values.st2.datastore_crypto_key }}
   datastore_crypto_key: {{ .Values.st2.datastore_crypto_key | b64enc }}
 {{- else }}
-  datastore_crypto_key: {{ tpl (.Values.default_datastore_crypto_key | toRawJson) . | b64enc }}
+  {{/* We template in Yaml to avoid excessive escaping of quotes in Json */}}
+  datastore_crypto_key: {{ tpl (.Values.default_datastore_crypto_key | toYaml) . | fromYaml | toRawJson | b64enc }}
 {{- end }}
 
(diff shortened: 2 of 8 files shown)


cognifloyd:
The last helm-e2e test used this to generate this:
```yaml
  datastore_crypto_key: eyJhZXNLZXlTdHJpbmciOiJTWGxSTElGYklNREtTL0ZrcEdIMG1aM3NmVHo4SVNGTGg3Qk04TkZENDZZIiwiaG1hY0tleSI6eyJobWFjS2V5U3RyaW5nIjoiVHNuYkEzOFI1WHdXdG9XMWRrTC1wcGlWVWY0SklnNmNkRTUtdFY3M1NNcyIsInNpemUiOjI1Nn0sIm1vZGUiOiJDQkMiLCJzaXplIjoyNTZ9
```

decoding that, we have:
```
  datastore_crypto_key: '{"aesKeyString":"SXlRLIFbIMDKS/FkpGH0mZ3sfTz8ISFLh7BM8NFD46Y","hmacKey":{"hmacKeyString":"TsnbA38R5XwWtoW1dkL-ppiVUf4JIg6cdE5-tV73SMs","size":256},"mode":"CBC","size":256}'
```

Which is perfectly valid.

eric-al (approved):
This looks great.  A bit of extra security by default with zero effort required!

conf/datastore_crypto_key.yaml
```diff
@@ -0,0 +1,13 @@
...
+# randBytes returns a base64 encoded string
+# 32 bytes = 256 bits / 8 bits/byte
+
+aesKeyString: '{{ randBytes 32 | replace "+" "-" | replace "_" "/" | replace "=" "" }}'
+mode: CBC
+size: 256
+
+hmacKey:
+  hmacKeyString: '{{ randBytes 32 | replace "+" "-" | replace "_" "/" | replace "=" "" }}'
+  size: 256
```
armab, line 13: Looks like it's making the impossible possible :) 

Overall I'm worried about the implementation and if that's a good way or not to try this in the chart really.
Even the slightest security risk behind the implementation/diff is sufficient to avoid the drill here generating the K/V crypto key.

I'd rely on someone better from the @StackStorm/tsc with security to review this. Maybe @punkrokk ?

values.yaml
```diff
@@ -49,7 +49,8 @@ st2:
   #password: Ch@ngeMe
   # ST2 crypto key for the  K/V datastore.
   # See https://docs.stackstorm.com/datastore.html#securing-secrets-admin-only for more info.
-  # Warning! Replace with your own generated key!
+  # If set, st2.datastore_crypto_key always overrides any existing datastore_crypto_key.
+  # If not set, the datastore_crypto_key is auto-generated on install and preserved across upgrades.
   #datastore_crypto_key: >-
   #  {"hmacKey": {"hmacKeyString": "", "size": 256}, "size": 256, "aesKeyString": "", "mode": "CBC"}
   # SSH private key for the 'stanley' system user ('system_user.ssh_key_file' in st2.conf)
```
armab, line 56: How to configure the chart to disable the use of the datastore crypto key?

cognifloyd: Why would you want to disable the datastore crypto key?

conf/datastore_crypto_key.yaml
cognifloyd: If there is a security issue with this formula, then st2-core would need to change, and every installation everywhere would have to update their datastore crypto key. This chart would, of course, also have to be updated to match whatever formula st2-core uses to generate these keys. I think the formula is easier to understand here than in st2-core, so it should be fairly simple to migrate this if ever needed in the future.

That said, I look forward to hearing what @punkrokk or other @StackStorm/TSC members have to say about this.

templates/secrets_datastore_crypto_key.yaml
```diff
@@ -2,12 +2,13 @@
 {{- $deprecated_crypto_key := (default (dict) (default (dict) .Values.secrets).st2).datastore_crypto_key }}
 {{- if $deprecated_crypto_key }}
 {{- fail "Please update your values! The datastore_crypto_key value moved from secrets.st2.* to st2.*" }}
-{{- else if .Values.st2.datastore_crypto_key }}
+{{- end }}
```
cognifloyd, line 5: If we need to support reducing security by disabling the datastore_crypto_key, then we could do something like:
```suggestion
{{- else if .Values.st2.datastore_crypto_key != "disable" }}
```
And do something similar everywhere else I've remoed this check.

cognifloyd:
Here is the st2 formula for generating the crypto key, trimmed down to make it easier to follow what is going on:

- The primary line of `st2-generate-symmetric-crypto-key` (note how `key_size=256` is hard-coded): [st2common/bin/st2-generate-symmetric-crypto-key#L49](https://github.com/StackStorm/st2/blob/b02617496807283b94ccd8be1b7e42b7677dbee1/st2common/bin/st2-generate-symmetric-crypto-key#L49)
  ```python
  from st2common.util.crypto import AESKey
  ...
  aes_key = AESKey.generate(key_size=256)
  ```
- `st2common.util.crypto.AESKey.generate`: [st2common/st2common/util/crypto.py#L133-L145](https://github.com/StackStorm/st2/blob/dfd533c5bfb0bcbdadcd539d39a6c97c244d35d8/st2common/st2common/util/crypto.py#L133-L145) and [#L431-L450](https://github.com/StackStorm/st2/blob/dfd533c5bfb0bcbdadcd539d39a6c97c244d35d8/st2common/st2common/util/crypto.py#L431-L450)
  ```python
  import base64
  ...
  DEFAULT_AES_KEY_SIZE = 256
  ...
  class AESKey(object):
      ...
      @classmethod
      def generate(self, key_size=DEFAULT_AES_KEY_SIZE):
          """
          Generate a new AES key with the corresponding HMAC key.
          ...
          """
          ...
          aes_key_bytes = os.urandom(int(key_size / 8))
          aes_key_string = Base64WSEncode(aes_key_bytes)

          hmac_key_bytes = os.urandom(int(key_size / 8))
          hmac_key_string = Base64WSEncode(hmac_key_bytes)

          return AESKey(
              aes_key_string=aes_key_string,
              hmac_key_string=hmac_key_string,
              hmac_key_size=key_size,
              mode="CBC",  # CBC is the only supported mode
              size=key_size,
          )


  def Base64WSEncode(s):
      """
      Return Base64 web safe encoding of s. Suppress padding characters (=).
      Uses URL-safe alphabet: - replaces +, _ replaces /. Will convert s of type
      unicode to string type first.
      ...
      NOTE: Taken from keyczar (Apache 2.0 license)
      """
      ...
      return base64.urlsafe_b64encode(s).decode("utf-8").replace("=", "")
  ```
- in standard library, `base64.urlsafe_b64encode`: https://github.com/python/cpython/blob/3.6/Lib/base64.py#L111-L118
  ```python
  _urlsafe_encode_translation = bytes.maketrans(b'+/', b'-_')
  ...
  def urlsafe_b64encode(s):
      """Encode bytes using the URL- and filesystem-safe Base64 alphabet.
      Argument s is a bytes-like object to encode.  The result is returned as a
      bytes object.  The alphabet uses '-' instead of '+' and '_' instead of
      '/'.
      """
      return b64encode(s).translate(_urlsafe_encode_translation)
  ```
- in standard library, `base64.b64encode`: https://github.com/python/cpython/blob/3.6/Lib/base64.py#L51-L62
  ```python
  def b64encode(s, altchars=None):
      """Encode the bytes-like object s using Base64 and return a bytes object.
      Optional altchars should be a byte string of length 2 which specifies an
      alternative alphabet for the '+' and '/' characters.  This allows an
      application to e.g. generate url or filesystem safe Base64 strings.
      """
      encoded = binascii.b2a_base64(s, newline=False)
      ...
      return encoded
  ```
- and finally: https://github.com/python/cpython/blob/3.6/Lib/base64.py#L108
  ```python
  _urlsafe_encode_translation = bytes.maketrans(b'+/', b'-_')  # ie replace '+' with '-'; replace '/' with '_'
  ```

cognifloyd:
And here is the implementation of `randBytes` which is a sprig template function available in helm templates:
https://github.com/Masterminds/sprig/blob/3ac42c7bc5e4be6aa534e036fb19dde4a996da2e/crypto.go#L70-L76
```go
import (
	...
	"crypto/rand"
	...
	"encoding/base64"
	...
)
func randBytes(count int) (string, error) {
	buf := make([]byte, count)
	if _, err := rand.Read(buf); err != nil {
		return "", err
	}
	return base64.StdEncoding.EncodeToString(buf), nil
}
```
This creates a buffer of the desired size, and then fills that buffer with random data, and then base64 enccodes it.

values.yaml
armab: That's how the previous configuration was before this PR.

cognifloyd: Would setting `datastore_crypto_key: disable` be ok as a way to disable this?

conf/datastore_crypto_key.yaml
armab: No worries. To be clear, I'm good with this new feature itself.

However, I'm looking for feedback if the key generation here is cryptographically secure. I didn't look what st2 does under the hood, but I'd trust those who good at this and if trying to mimic the st2 behavior for st2-krypto-key-generation in the Helm template engine is good enough. 

I think the folks would need to dig deeper into the code you provided :+1:

values.yaml
armab: I'm looking at the `st2.conf` example  https://github.com/StackStorm/st2/blob/master/conf/st2.conf.sample#L176-L177
If the `encryption_key_path =` is empty, stackstorm just won't use it.

I think if we'll follow that way it would be simple enough without an additional setting.
Would also make sense to document the empty option in the Helm values comments.

punkrokk:
My only comment is we could probably use AES512. 

I think the function to create the key is fine.

cognifloyd:
If we add support for AES512 in StackStorm itself, them we can update this to use the better encryption by default.

values.yaml
cognifloyd: If `encryption_key_path` is not defined in st2.conf, won't StackStorm refuse to encrypt any datastore entries? ie nothing would have to be re-encrypted since nothing CAN be encrypted without an encryption key. So, I don't think adding an encryption key to existing installations (where none is currently provided) would be a problem. Do I have a faulty assumption somewhere in here?

As far as making it possible to disable the encryption key: This PR auto-generates the datastore_crypto_key if it is not defined. How can I distinguish between the new default (datastore_crypto_key value is empty so auto-generate it) and the old default (default_crypto_key value is empty, so set encryption_key_path to empty as well). Maybe I should flag this as a breaking change in the changelog? I think this is backwards compatible, as described in the last paragraph.

cognifloyd: distinguishing between `# datastore_crypto_key` (unset) and `datastore_crypto_key: ""` (empty string) makes the template more complex and feels hacky: https://github.com/Masterminds/sprig/issues/53#issuecomment-483414063

It's complex enough that I'd probably need to introduce a helper just to determine whether or not to create or mount the crypto_key secret.

armab: Thanks for researching this, the new flag would be nice then per your initial idea :+1: 

And yes, in any way this is a breaking change as people who had no st2 K/V crypto key configured in Helm values would find that it'll be forcefully autogenerated for them after the upgrade, including the change in st2.conf.

cognifloyd: done

armab (approved):
Thanks for the PR @cognifloyd, and @punkrokk for additional review :+1:

PR #267, opened 2021-11-16
Added https port to st2web service if ssl = true

sandesvitor:
## Description

Adding https port to st2web service if .Values.st2web.service.ssl equals `true`. .Values.st2web.service.ssl defaults to `false` for backward compatibility.


cognifloyd:
I did something similar in #264. Do you need to expose both http and https? Or would #264 meet your needs?

sandesvitor:
Hey @cognifloyd ! Yeah, #264 meet our needs =)

Thanks! May I close this PR?

cognifloyd:
Cool. Let's leave this open until @armab has time to take a look. I don't know which approach (something like this one or #264) he would prefer.

cognifloyd:
Closing this in favor of #264 which Eugen approved and I merged.

PR #268, opened 2021-11-17
fix st2web.postStartScript indent

cognifloyd:
The indent for `st2web-config-vol` and `st2-post-start-script-vol` were not aligned. If both are provided, the generated yaml will fail to parse.

So, this fixes the indent.


armab (approved):
Good catch!

PR #270, opened 2021-12-02
apply annotations to deployments and jobs, not just pods

cognifloyd:
Apply annotations (added to Pods in #195) to Deployments and Jobs.

I need to apply annotations to Deployments due to cluster policies that require annotations on Deployments and not just on Pods.


armab (approved):
:+1:

PR #276, opened 2021-12-21
Extra volumes for logging conf

### LINKED ISSUE #269: Feature Request: Allow configuration of logging to GELF and/or logfmt + custom formats
Basically, expose the python logging configuration in the helm chart and provide pre-canned examples for popular formats

cognifloyd:
The logging conf files are baked into the docker image and provide only the minimal logging level. Generally this config can be updated by building a new docker image.

However, in some cases, it is helpful to be able to turn on/off debug logging to diagnose issues occurring only in particular clusters or environments. Sadly, there is no way to use env vars to configure that in StackStorm, so I need something more flexible to update logging conf files.

So, this expands the `extra_volumes` feature so that all of the python services can use it. Then, to update the logging conf file, just put the custom logging conf in a custom ConfigMap (managed outside of this chart), and then load it into the relevant deployments using `extra_volumes` values.

Resolves #269


cognifloyd:
This resolves #269

cognifloyd:
Rebased. @armab This was requested by someone else in #269. Can you review this please?

armab (approved):
Good stuff, thanks! :+1:

PR #282, opened 2022-02-13
feat: add CHANGELOG entry for container version bump to Python 3.8 / …

jstaph:
Adds CHANGELOG line to note future bump to Python 3.8 / Ubuntu 20


armab (changes requested):
Thanks!

Found a minor thing.
Could you please also update the tests here for the new py version?
https://github.com/StackStorm/stackstorm-ha/blob/62b33259bb72c60801ba8326c435276289b6b51c/tests/st2tests.sh#L7-L13

armab (approved):
The tests should pass once we release the `v3.7.0` + updated Docker images.

Thanks for the help preparing it! :+1:

jstaph:
You bet, and nice catch on the tests!  Much appreciated!

cognifloyd (approved):

PR #283, opened 2022-02-16
Basic Unit Test Infrastructure

cognifloyd:
This is part of #28.

Using `pytest`, this introduces a way to unit test the rendered templates.

For now, there is only one test showing how the infra works.
After this is merged, we can start adding tests for other template features.

### Why not helm-unittest?

These links explain why apache/airflow decided to abandon helm-unittest in favor of a pytest-based chart testing approach:

- https://github.com/helm-unittest/helm-unittest/issues/110
- https://github.com/apache/airflow/issues/11657

In short, I agree with their logic. StackStorm is a python-based solution. Helm offers some integration-style testing apparatus, but unit tests are rather difficult.
Different people have been maintaining helm-unittest, so there is no one canonical "upstream" repo for it. You have to search and figure out which one is the most maintained.
Plus, the YAML-based tests will be something else to learn and using python to test should allow for more expressive tests.

### Why pytest?

nosetest is not well maintained and is less well known. Eventually I want to replace nosetest with pytest in the main st2 repo. I've done quite a bit of preparatory work in that direction.
I do not want to introduce yet another unittest dependecy that will have to be migrated in the future.

I also strongly dislike the unittest-style `self.assert*` and I find debugging unittest tests very annoying.
Pytest-enhanced raw asserts are much more straightforward.

apache/airflow has an odd combo of unittest classes that they run with pytest. I avoid Unittest wherever possible, so I created pytest fixtures in this PR.

### tests/helm_template_generator.py

tests/helm_template_generator.py was developed by apache/airflow.
That project is Apache 2.0 licensed, so we can reuse it here.

To maintain authorship data, I imported the git history (using [this](https://www.pixelite.co.nz/article/extracting-file-folder-from-git-repository-with-full-git-history/)) for this file:
https://github.com/apache/airflow/blob/main/chart/tests/helm_template_generator.py
I also included the LICENSE file to make sure licensing is clear in our chart repo, and a copy of the NOTICE file from the apache/airflow repo.

### DIFF
--- a/tests/test_annotations.py
+++ b/tests/test_annotations.py
@@ -13,7 +13,9 @@ import pytest
         ),
     ],
 )
-def test_annotations(chart_resources: List[Dict[str, Any]], expected_annotations: Dict[str, str]) -> None:
+def test_annotations(
+    chart_resources: List[Dict[str, Any]], expected_annotations: Dict[str, str]
+) -> None:
     """Make sure that annotations are working"""
     # based on apache/airflow:chart/tests/tests_annotations::test_annotations_are_added
 
--- a/tests/test_annotations.py
+++ b/tests/test_annotations.py
@@ -1,0 +1,26 @@
+from typing import Any, Dict, List
+
+import pytest
+
+
+@pytest.mark.parametrize(
+    ("values", "show_only", "expected_annotations"),
+    [
+        (
+            {"serviceAccount": {"serviceAccountAnnotations": {"foo": "bar"}}},
+            ["templates/service-account.yaml"],
+            {"foo": "bar"},
+        ),
+    ],
+)
+def test_annotations(chart_resources: List[Dict[str, Any]], expected_annotations: Dict[str, str]) -> None:
+    """Make sure that annotations are working"""
+    # based on apache/airflow:chart/tests/tests_annotations::test_annotations_are_added
+
+    # there's only one resource because we sepecified show_only.
+    assert len(chart_resources) == 1
+    service_account = chart_resources[0]
+
+    for name, value in expected_annotations.items():
+        assert name in service_account["metadata"]["annotations"]
+        assert value == service_account["metadata"]["annotations"][name]
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -1,0 +1,31 @@
+---
+name: Python package
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10"]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install pytest
+    - name: Test with pytest
+      run: |
+        pytest
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -3,9 +3,12 @@ name: Python package
 
 on:
   push:
-    branches: [ master ]
+    branches:
+      - master
+      - unit-tests  # TODO: drop this
   pull_request:
-    branches: [ master ]
+    branches:
+      - master
 
 jobs:
   build:
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -1,5 +1,5 @@
 ---
-name: Python package
+name: Unit Tests
 
 on:
   push:
@@ -10,7 +10,7 @@ on:
       - master
 
 jobs:
-  build:
+  pytest:
 
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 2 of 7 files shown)


cognifloyd:
Please note: I added 100 lines in python files that I authored. I also added 15 of the lines in NOTICE and all 49 lines of the gha workflow. The other 374 lines (201 in the Apache 2.0 LICENSE file) were copied from apache/airflow.

So, if the pull-request-size bot only evaluated the lines I wrote, this would be a `size/L` PR, not `size/XL`.

tests/test_annotations.py
```diff
@@ -0,0 +1,28 @@
...
+    """Make sure that annotations are working"""
+    # based on apache/airflow:chart/tests/tests_annotations::test_annotations_are_added
+
+    # there's only one resource because we sepecified show_only.
+    assert len(chart_resources) == 1
+    service_account = chart_resources[0]
+
+    for name, value in expected_annotations.items():
+        assert name in service_account["metadata"]["annotations"]
+        assert value == service_account["metadata"]["annotations"][name]
```
armab, line 28: I had experience working with somewhat similar Unit and Integration Tests written for K8s in another project for the LF exams.

While it looks powerful at the first though, I don't think it's the best fit for a simple project like `stackstorm-ha`.
The disadvantage to me is spaghetti and confusion when debugging, reading these tests, maintaining them or even asking someone to create a test. The amount of indirection and complexity is excessive.


At this point, I'd better go with the simplest possible approach. Even doing helm template, `yq` or `jq` with `jsonpath` that could be ran from the CLI would be a better, when it's much easier to debug and connect the dots from Helm template itself to the Helm template test.

Yaml to Yaml is simple as that.

armab (changes requested):
Thanks for putting the experiment together and ideas here!

While I understand the approach Apache Airflow has chosen, I believe they're driven by different needs. Just quickly take a look at the https://github.com/apache/airflow/tree/main/chart/templates and development velocity.

A somewhat similar approach would be going with the https://github.com/instrumenta/helm-conftest + https://github.com/open-policy-agent/conftest via Open Policy Agent (https://www.openpolicyagent.org/). That would be a true cloud-native CNCF approach. In the Airflow PR, they mention [terratest](https://github.com/gruntwork-io/terratest-helm-testing-example/blob/master/test/minimal_pod_helm_template_engine_integration_test.go) which is relevant to this approach, too.

However, both `golang` and `python` are overkill for this specific `stackstorm-ha` project.

Let's compare these 2 PRs:
- https://github.com/jenkinsci/helm-charts/pull/451/files
- https://github.com/apache/airflow/pull/21484/files

Adding a testing case for the templates shouldn't be harder than writing the templates themselves.

I believe, our situation is that we have different needs and priorities:
- Low development velocity. After stabilizing the helm chart there will be very few things to add.
- Simplicity & Maintainability
- No really complex scenarios
- Creating unit tests should be easy as possible
  - show by example (a copy-paste from the other file)
  - applies to external contributors
- Adding unit tests is a bonus. The chart is already successful in its state, even if we don't add the unit testing framework here.

And so navigating all the unit test infrastructure after, for example, 6 months of not touching this codebase should be straightforward. Requesting a contributor to add a quick unit test based on the examples we have should be easy as copy-pasting a couple of lines.
We don't need all the possible flexibility. We need to really add a few YAML lines in deployment and connect them with a few respective YAML lines in the tests. Mentally, it's very easy and does the job.

https://github.com/quintush/helm-unittest is known as a working fork for `helm unittest` plugin in the K8s community, - that should work fine. See https://github.com/quintush/helm-unittest/blob/master/test/data/v3/basic/tests/ and https://github.com/jenkinsci/helm-charts/tree/main/charts/jenkins/unittests/. They provide a good overview of examples.

armab:
Found where I saw the `BATS` unit tests before with simplest approach like `helm template | yq`.
It's Hashicorp: https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/test/unit/dns-service.bats

The advantage I think there is no need to leave the CLI and easy to reproduce and debug. Connecting dots from the Helm template to the test results is easy and so the room for error is lower.  And we're already using BATS for [integration & smoke](https://github.com/StackStorm/stackstorm-ha/blob/master/tests/st2tests.sh) testing.
^^ That's `level 0`.

https://github.com/quintush/helm-unittest represents things a bit nicer and more user-friendly with YAML. That's `level 1`.

Python vs Golang unit tests would be `level 2` in the cases when flexibility and requirements for the testing coverage with edge cases are non-negotiable.

tests/test_annotations.py
cognifloyd: Doing things from the CLI is not simpler! It's a royal pain to maintain tests in bash and I hate writing tests that way.

cognifloyd:
> Low development velocity. After stabilizing the helm chart there will be very few things to add.

Yes. This is a big problem. We MUST speed up the velocity. It is far too slow and takes far too long to get things reviewed and merged. The slower it is the less community engagement we can get.

cognifloyd:
> Adding unit tests is a bonus. The chart is already successful in its state, even if we don't add the unit testing framework here.

Nope. I can't get the features I'm trying to add merged until you're confident everything is covered by unit tests. It is not a bonus. We need it so we can speed up velocity.

tests/test_annotations.py
cognifloyd: So, yet again, you and I have different definitions of simple.

cognifloyd:
:bow: I opened #284 to replace this one.

PR #284, opened 2022-02-17
Add helm-unittest

cognifloyd:
Per #28 and https://github.com/StackStorm/stackstorm-ha/pull/283#pullrequestreview-885020923 unittests will use helm-unittest.

This adds a basic GHA workflow and a single dummy test. The dummy test is commented to document the various options in the test file.

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -1,0 +1,45 @@
+---
+name: Unit Tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  helm-unittest:
+
+    runs-on: ubuntu-latest
+#    strategy:
+#      matrix:
+        # Relevant tools installed by default on ubuntu 20.04:
+        #   - helm 3.8.0
+        #   - jq 1.6
+        #   - kind 0.11.1
+        #   - kubectl 1.23.3
+        #   - minikube 1.25.1
+        #   - python 3.8.10
+        #   - yamllint 1.26.3
+        #   - yq 4.19.1
+        # see: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install helm-unittest
+      # We should periodically check to see if another fork has taken over maintenance,
+      # as the de-facto "best" fork has changed several times over the years.
+      run: |
+        helm plugin install https://github.com/quintush/helm-unittest
+
+    - name: Install chart dependencies
+      run: |
+        helm dependency update
+
+    - name: Run helm-unittest
+      # by default looks for tests/*_test.yaml
+      run: |
+        helm unittest --color --helm3 .
--- a/tests/dummy_test.yaml
+++ b/tests/dummy_test.yaml
@@ -1,0 +1,76 @@
+---
+suite: Dummy Tests
+templates:
+  - configmaps_st2-conf.yaml
+
+# this optionally overrides {{ .Release }}
+#release:
+#  name: my-release         # default: release-name
+#  namespace: my-namespace  # default: namespace
+#  revision: 1              # default: 0
+#  isUpgrade: true          # default: false
+
+# this optionally overrides {{ .Capabilities }}
+#capabilities:
+#  majorVersion: 1          # default from Helm
+#  minorVersion: 10         # default from Helm
+#  apiVersions:             # defaults to versionset used by k8s version
+#    - br.dev.local/v2
+
+# this optionally overrides {{ .Chart }}
+#chart:
+#  version: 1.0.0           # default from Chart.yaml
+#  appVersion: 1.0.0        # default from Chart.yaml
+  
+tests:
+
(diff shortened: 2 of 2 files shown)


armab (commented):
Thanks for composing a quick example, looks really sleek! :+1: 

Do we want to split the directory structure here, considering we now have unit vs integration tests?
https://github.com/StackStorm/stackstorm-ha/tree/master/tests
Adding a new Readme.md for Unit tests would be helpful as well.

I'd welcome the changelog as that'll help other developers seeing unit test framework addition as another stability message.

tests/README.md
```diff
@@ -18,4 +47,4 @@ To show the test results:
 kubectl logs <release-name>-st2tests
 ```
 
-See https://helm.sh/docs/developing_charts/#chart-tests with more information about Helm chart tests.
+See https://helm.sh/docs/topics/chart-tests/ with more information about Helm chart tests.
```
cognifloyd, line 50: old link was to helm v2 docs. This is the same doc but for helm v3.

tests/README.md
```diff
@@ -1,6 +1,35 @@
...
+helm dependency update
+```
+
+To run the tests manually from the chart's root dir:
+```
+helm unittest --helm3 tests/unit/*_test.yaml .
+```
+
+Note that `helm-unittest` still defaults to helm 2, so you must pass `--helm3` or `-3` for short.
+You must also tell it where the unit tests are with `-f` because we keep unit tests in the `tests/unit` directory.
```
cognifloyd, line 24: OK. I moved the tests to `tests/unit`.
That means we have to tell `helm-unittest` where to find the test files (by default it looks for `tests/*_test.yaml`).
Sadly `helm-unittest` has no config file so every invocation will have to specify which files to test with.

cognifloyd: @armab Can we just keep the unit tests directly under `tests/`? If we don't I'm unlikely to run them locally because specifying the path every time would be annoying. If we leave them in `tests/unit` I'll just let CI run the tests for me.

cognifloyd: I moved them back to `tests/` and documented the file name scheme: `tests/unit-name_your_test.yaml`.
This will keep the unit tests sorted within the `tests/` directory without making it awkward to run the unit tests.

cognifloyd:
I added tests for custom annotations and removed the dummy unit test.

tests/README.md
```diff
@@ -1,6 +1,40 @@
...
+To run the tests manually from the chart's root dir:
+```
+helm unittest --helm3 .
+```
+
+Note that `helm-unittest` still defaults to helm 2, so you must pass `--helm3` or `-3` for short.
+`helm-unittest` does not have a configuraiton file which makes using non-standard directories (like `tests/unit/`)
+very inconvenient (you would have to specify `-f tests/unit/*_test.yaml` on every invocation).
+So, we keep all of the unit test files under tests, but prefix them with `unit-` to separate them from
+the integration tests.
```
armab, line 27:
```suggestion
```
Remove. It may sound a bit like a complaint, rather than documentation, which is not what we need in the docs 🙂 Just providing a CLI example was sufficient.

armab (changes requested):
No problem with running `helm unittest --helm3 -f tests/unit/*_test.yaml` to structure the tests in `unit/` and `integration/` dirs.

Looks like the others were successful in this already: 
https://github.com/jenkinsci/helm-charts/blob/26affd02465f7ef444967b0528142a6cd6b9642a/ct.yaml#L8

Let's please do something similar.

I understand it comes with a price of adding another `-f tests/unit/*_test.yaml` CLI param, but the other side is that the structure would be more obvious and predictable for everyone.

tests/unit-custom_annotations_test.yaml
```diff
@@ -0,0 +1,169 @@
...
+        service:
+          hostname: some-host-name
+          annotations: *annotations
+    asserts: *annotations_asserts
+
+  # st2chatops service does not accept custom annotations
+
+  - it: Deployments+Pods accept custom annotations
+    template: deployments.yaml
+    set:
```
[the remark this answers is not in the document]
armab: Nice :+1:

tests/README.md
armab: I think another CLI param is a reasonable price to pay to structure them in a usual way. Replied below.

cognifloyd: done

tests/README.md
```diff
@@ -1,6 +1,34 @@
...
+helm plugin install https://github.com/quintush/helm-unittest
+helm dependency update
+```
+
+To run the tests manually from the chart's root dir:
+```
+helm unittest --helm3 -f tests/unit/*_test.yaml .
+```
+
+> Note! If you need to add unit tests, file names should follow this pattern: `tests/unit-name_your_test.yaml`
```
armab, line 23:
```suggestion
> Note! If you need to add unit tests, file names should follow this pattern: `tests/unit/name_your_test.yaml`
```

armab (approved):
Big thanks for all the changes! :+1: 

So I understand the groundwork here will help you to ship the https://github.com/StackStorm/stackstorm-ha/pull/218. 
 +100 on the approach.

PR #289, opened 2022-02-25
Add Ability to Write/Use st2.secrets.conf from K8s Secret

eric-al:
This PR adds the capability to define '/etc/st2/st2.secret.conf' in a Kubernetes secret.  No longer will users need things like helm-secrets to safely store secret values that ultimately wind up in /etc/st2/*.conf.

If this is defined, the contents of key 'st2.secret.conf' within the defined secret name will be written to '/etc/st2/st2.secret.conf' on all pods.  In addition, '--config-file=/etc/st2/st2.secrets.conf' will be added to the end of the command line arguments for all pods, superseding all other configuration values.

This addresses the Comment-Based-TODO here:  https://github.com/alertlogic/stackstorm-ha/blob/master/templates/configmaps_st2-conf.yaml#L16

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Prefix template helpers with chart name and format helper comments as template comments. (#272) (by @cognifloyd)
 * New feature: Add `extra_volumes` to all python-based st2 deployments. This can facilitate changing log levels by loading logging conf file(s) from a custom ConfigMap. (#276) (by @cognifloyd)
 * Initialize basic unittest infrastructure using `helm-unittest`. Added tests for custom annotations. (#284)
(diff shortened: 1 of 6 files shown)


sanketgawade (approved):
+1

cognifloyd:
Could you add a unit test that inspects the command? You can see other unit tests in `tests/unit/`.
The test should probably ensure that `command` is defined on all deployments and jobs and that it includes the expected args.

cognifloyd (changes requested):
As this PR is made from an organization repo, github doesn't allow you to give us permission to update the branch. So, I can't merge in master or do any cleanups.

Would you please:
- merge in master
- minor cleanup in comments

For the unit tests, I believe something like this should work: https://gist.github.com/cognifloyd/362a3637e43e199a013897f74454c14e

I haven't tested it, but I think that'll cover all of the testing boilerplate around these conf files.

cognifloyd:
Closing in favor of #319 since Eric no longer has access to the @eric-al account and we as maintainers cannot edit this PR to help him with it.

PR #291, opened 2022-03-03
Allow "external" st2api, st2auth, and st2stream Services (LoadBalancer type with hostname)

sandesvitor:
## Description

New feature to include possibility for external services in st2api and st2auth, setting default value for this services as `ClusterIP` and `hostname: ""`.

## Context

To enable remote servers remote communication with stackstorm, and not just within Kubernetes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * Prefix template helpers with chart name and format helper comments as template comments. (#272) (by @cognifloyd)
 * New feature: Add `extra_volumes` to all python-based st2 deployments. This can facilitate changing log levels by loading logging conf file(s) from a custom ConfigMap. (#276) (by @cognifloyd)
 * Initialize basic unittest infrastructure using `helm-unittest`. Added tests for labels, custom annotations, SecurityContext, pullSecrets, pullPolicy, Resources, nodeSelector, tolerations, affinity, dnsPolicy, dnsConfig, ServiceAccount attach, postStartScript, and both sensor-modes. (#284, #288)
-* New feature to include possibility for external services in st2api and st2auth, setting default value for this services as `ClusterIP` and `hostname: ""`. (by @sandesvitor)
+* New feature to include possibility for external services in st2api, st2stream and st2auth, setting default value for this services as `ClusterIP` and `hostname: ""`. (by @sandesvitor)
 
 ## v0.80.0
 * Switch st2 to `v3.6` as a new default stable version (#274)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * New feature: Add `extra_volumes` to all python-based st2 deployments. This can facilitate changing log levels by loading logging conf file(s) from a custom ConfigMap. (#276) (by @cognifloyd)
 * Initialize basic unittest infrastructure using `helm-unittest`. Added tests for labels, custom annotations, SecurityContext, pullSecrets, pullPolicy, Resources, nodeSelector, tolerations, affinity, dnsPolicy, dnsConfig, ServiceAccount attach, postStartScript, and both sensor-modes. (#284, #288)
 * Allow partitioning sensors using the hash_range strategy instead of one sensor per pod. (#218) (by @cognifloyd)
-* New feature to include possibility for external services in st2api, st2stream and st2auth, setting default value for this services as `ClusterIP` and `hostname: ""`. (by @sandesvitor)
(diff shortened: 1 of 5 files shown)


cognifloyd:
Would you please add some unit tests for the service `type` and `externalName`? And maybe for the annotation as well?
You can see examples tests in `tests/unit/` (most of which deal only with deployments and jobs - the service tests will be mock more straight forward).

Also, `st2stream` should have the same options as `st2auth` and `st2api`. Would you please add these changes to `st2stream` as well?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Prefix template helpers with chart name and format helper comments as template comments. (#272) (by @cognifloyd)
 * New feature: Add `extra_volumes` to all python-based st2 deployments. This can facilitate changing log levels by loading logging conf file(s) from a custom ConfigMap. (#276) (by @cognifloyd)
 * Initialize basic unittest infrastructure using `helm-unittest`. Added tests for labels, custom annotations, SecurityContext, pullSecrets, pullPolicy, Resources, nodeSelector, tolerations, affinity, dnsPolicy, dnsConfig, ServiceAccount attach, postStartScript, and both sensor-modes. (#284, #288)
+* New feature to include possibility for external services in st2api and st2auth, setting default value for this services as `ClusterIP` and `hostname: ""`. (by @sandesvitor)
```
armab, line 8: Looks like we'll also need similar functionality for `st2stream` so users can configure all the st2 3 services in the same consistent way via Helm values.

tests/unit/custom_annotations_test.yaml
```diff
@@ -52,14 +52,26 @@ tests:
 
   # st2auth, st2api, st2stream services do not accept custom annotations
 
-  - it: st2web Service accepts custom annotations
+  - it: st2web, st2auth, st2api ans st2stream Services accepts custom annotations
     template: services.yaml
     documentIndex: 3
```
cognifloyd, line 57: Remove this line to test more than just st2web.
```suggestion
```

tests/unit/custom_annotations_test.yaml
```diff
@@ -52,14 +52,26 @@ tests:
 
   # st2auth, st2api, st2stream services do not accept custom annotations
 
-  - it: st2web Service accepts custom annotations
+  - it: st2web, st2auth, st2api ans st2stream Services accepts custom annotations
```
cognifloyd, line 55:
```suggestion
  - it: st2web, st2auth, st2api and st2stream Services accepts custom annotations
```

tests/unit/custom_annotations_test.yaml
```diff
@@ -52,14 +52,26 @@ tests:
...
+          hostname: some-host-name
+          annotations: *annotations
+      st2auth:
+        service:
+          hostname: some-host-name
+          annotations: *annotations
+      st2stream:
+        service:
+          hostname: some-host-name
+          annotations: *annotations
```
cognifloyd, line 74: removing the `documentIndex` line will make it test all of the services in the file.
The only service we don't want to test here is `st2chatops` since it does not have the same annotations.
Though `st2chatops` is disabled by default, I prefer to make the values explicit that I'm relying on for the test like this:
```suggestion
          annotations: *annotations
      st2chatops:
        enabled: false # exclude st2chatops to only test other services
```

cognifloyd (commented):
Excellent. Just a few comments to help adjust the annotations test.

sandesvitor:
Thanks for the tips, @cognifloyd ! But I confess I'm struggling with helm-unittest hehe mostly this new file that I created, tests/unit/services_test.yaml

tests/unit/services_test.yaml
```diff
@@ -0,0 +1,20 @@
...
+        service:
+          hostname: some-host-name
+          type: LoadBalancer
+    asserts:
+      - equal:
+        path: spec.type
+        value: LoadBalancer
+      - equal:
+        path: spec.externalName
+        value: some-host-name
```
cognifloyd, line 20: You're missing a bit of indentation:
```suggestion
      - equal:
          path: spec.type
          value: LoadBalancer
      - equal:
          path: spec.externalName
          value: some-host-name
```

cognifloyd (commented):
I'm happy to help. It took my awhile to figure out helm-unittest as well.
While writing tests, I like to keep this page open:
https://github.com/quintush/helm-unittest/blob/master/DOCUMENT.md

tests/unit/services_test.yaml
sandesvitor: Oh... hehhe sorry! Fixing it =)

cognifloyd (approved):

PR #292, opened 2022-03-05
Add unit tests for env, envFrom, st2.packs.images, and st2.packs.volumes

cognifloyd:
- add tests for env and envFrom
- test deployments and jobs with/out st2.packs.images and without st2.packs.volumes
- test deployments and jobs with/out st2.packs.images and with st2.packs.volumes
- test misconfigured st2.packs.volumes

### DIFF
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -485,7 +485,6 @@ spec:
         envFrom:
         - configMapRef:
             name: {{ .Release.Name }}-st2-urls
-            optional: true
     {{- if or .Values.st2web.config .Values.st2web.extra_volumes .Values.st2web.postStartScript }}
         volumeMounts:
     {{- else }}
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -1964,6 +1964,9 @@ spec:
           configMap:
             name: {{ .Release.Name }}-st2chatops-post-start-script
       {{- end }}
+    {{- else }}
+        volumeMounts: []
+      volumes: []
     {{- end }}
     {{- if .Values.st2chatops.serviceAccount.attach }}
       serviceAccountName: {{ template "stackstorm-ha.serviceAccountName" . }}
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -1306,7 +1306,7 @@ spec:
           {{- end }}
         {{- end }}
         {{- if $sensor.env }}
-        env: {{- include "stackstorm-ha.customEnv" $ | nindent 8 }}
+        env: {{- include "stackstorm-ha.customEnv" $sensor | nindent 8 }}
         {{- end }}
         envFrom:
         - configMapRef:
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -1306,7 +1306,7 @@ spec:
           {{- end }}
         {{- end }}
         {{- if $sensor.env }}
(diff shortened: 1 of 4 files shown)


templates/deployments.yaml
```diff
@@ -1306,7 +1306,7 @@ spec:
           {{- end }}
         {{- end }}
         {{- if $sensor.env }}
-        env: {{- include "stackstorm-ha.customEnv" . | nindent 8 }}
+        env: {{- include "stackstorm-ha.customEnv" $ | nindent 8 }}
```
cognifloyd, line 1309: This is a bug uncovered by the env tests.

templates/deployments.yaml
```diff
@@ -1964,6 +1964,9 @@ spec:
           configMap:
             name: {{ .Release.Name }}-st2chatops-post-start-script
       {{- end }}
+    {{- else }}
+        volumeMounts: []
+      volumes: []
```
cognifloyd, line 1969: This simplified a chunk of the (already lengthy) volume tests.

templates/deployments.yaml
```diff
@@ -1306,7 +1306,7 @@ spec:
           {{- end }}
         {{- end }}
         {{- if $sensor.env }}
-        env: {{- include "stackstorm-ha.customEnv" . | nindent 8 }}
+        env: {{- include "stackstorm-ha.customEnv" $sensor | nindent 8 }}
```
cognifloyd, line 1309: The env test caught this typo

tests/unit/env_test.yaml
```diff
@@ -0,0 +1,252 @@
...
+      - contains: *contains_st2_urls
+        documentIndex: 1
+      - contains: *contains_st2_urls
+        documentIndex: 2
+      - contains:
+          path: spec.template.spec.containers[0].envFrom
+          content:
+            configMapRef:
+              name: st2ha-st2-urls
+              optional: true # TODO: consider dropping this unique diff from other deployments
```
cognifloyd, line 156: We might want to drop this `optional: true` line from the st2web deployment as it makes the tests more complicated.
@armab It's been a long time, but do you remember why this was marked optional for st2web but not for anything else?

cognifloyd:
@armab could you review this? It may be large, but that's because the unit tests are very repetitive to cover the various possibilities (`helm-unittest` doesn't have anything like pytest's `parametrize`).

tests/unit/env_test.yaml
armab: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/
> #### Optional References
> A ConfigMap reference may be marked "optional". If the ConfigMap is non-existent, the mounted volume will be empty. If the ConfigMap exists, but the referenced key is non-existent the path will be absent beneath the mount point.
>
> You must create a ConfigMap before referencing it in a Pod specification (unless you mark the ConfigMap as "optional").

No, don't remember why would we need it.
Maybe something was missed during the enterprise -> oss transition.

I'm good dropping it :+1:

armab (approved):
Thanks!

tests/unit/env_test.yaml
cognifloyd: dropped.

cognifloyd:
Thanks for reviewing! I know this was pretty large.

PR #301, opened 2022-04-10
Ingress api version deprecation support for k8s v.1.22. fixes #277

### LINKED ISSUE #277: Support for Kubernetes 1.22 - networking.k8s.io/v1beta1 removed
Several APIs where deprecated in Kubernetes 1.22, in particular the `networking.k8s.io/v1beta1` (used [here](https://github.com/StackStorm/stackstorm-ha/blob/7d55c4964ef33893e042ba84e1756f1650172d39/templates/ingress.yaml#L3)). This prevent `ingress.yaml` to deploy successfully on Kubernetes 1.22.

I think it could be possible to support both APIs, using conditional logic ([Grafana does this](https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/_helpers.tpl#L132-L163), for example).

arms11:
`Ingress.yaml` uses correct `apiVersion` and related `spec` changes that are now required starting version `v1.22` of `kubernetes`.

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -1,6 +1,10 @@
 {{- if .Values.ingress.enabled }}
 ---
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+apiVersion: networking.k8s.io/v1
+{{- else }}
 apiVersion: networking.k8s.io/v1beta1
+{{- end }}
 kind: Ingress
 metadata:
   name: {{ .Release.Name }}-st2web-ingress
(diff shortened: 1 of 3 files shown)


cognifloyd:
Please add a changelog. And would you mind adding some unit tests for the Ingress? There are examples in tests/unit and see the readme under tests

arms11:
Thanks @cognifloyd for the feedback. Will take a look tomorrow.

cognifloyd (approved):
K. I added the changelog entry and tests.

templates/ingress.yaml
```diff
@@ -20,9 +24,18 @@ spec:
       paths:
       {{- range .paths }}
         - path: {{ default "/*" .path }}
+          {{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```
cognifloyd, line 27: The tests uncovered a bug here. It was missing `$` so I added that.

PR #302, opened 2022-04-11
Add custom pypi feature

chris3081:
This feature allows the consumer to specify a custom pypi repository.

Use cases:
 - Air gapped installs
 - Shared volume pack installs that use a private repo.

### DIFF
--- a/conf/pip.conf
+++ b/conf/pip.conf
@@ -1,0 +1,4 @@
+[global]
+index-url = https://pypi.org/simple
+
+extra-index-url = {{ .Values.st2.customPyPiRepUrl }}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -298,3 +298,26 @@ Create the custom env list for each deployment
   value: {{ $value | quote }}
   {{- end }}
 {{- end -}}
+
+{{/*
+Define pypi volumeMounts
+*/}}
+{{- define "stackstorm-pypi-volume-mounts" -}}
+{{- if .Values.st2.customPyPiRepUrl }}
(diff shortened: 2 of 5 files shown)


cognifloyd:
Have you seen the `extra_volumes` feature? Would that meet your needs here?

See https://github.com/StackStorm/stackstorm-k8s/blob/master/values.yaml#L644

chris3081:
This didn't look to entirely fir the use case, trying to think how this could be used to easily create the pip conf secret. Happy to go that route and drop the PR, I was just trying to add some that didn't need an external secret to the helm chart.

armab:
I agree with @cognifloyd.

From the chart design, we want to ensure if we could provide some generic approach that would work for many use cases, not just custom pip file. Looks like existing extra volumes/mounts might solve your needs, @chris3081 but please try and let us know.

It's also a common practice to create a secret outside of the Helm chart and it's used in the other community charts.
I'd recommend looking at https://github.com/StackStorm/stackstorm-k8s#extending-this-chart that'll help simplify your approach with managing custom external objects like secrets and other resources.

cognifloyd:
I use the [deliveryhero/k8s-resources](https://artifacthub.io/packages/helm/deliveryhero/k8s-resources) chart to manage the custom secrets that I need for StackStorm. Hope that helps!

Using `extra_volumes` for `/etc/pip.conf` is a usecase I hadn't considered. So, maybe you could add another example in the values.yaml  section I linked to earlier.

Also, would a section in the readme have helped you find out about `extra_volumes`?

chris3081:
Ok no problem

PR #316, opened 2022-06-07
Bump the version of Redis chart to the latest

anrajme:
Raising this PR as per [305](https://github.com/StackStorm/stackstorm-k8s/issues/305 ) - validating the E2E testing for the Redis version bump. As per the release notes [here](https://github.com/bitnami/charts/tree/master/bitnami/redis#to-1400), version 14.0.0 may have breaking changes for older installs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Fix bug that hung an init container when `st2.packs.volumes.enabled` without `st2.packs.volumes.configs`. (#324) (by @rebrowning)
 * Add ability to create custom labels for service account.(#327)(by @SuganJoe)
 * Fix bug that would not set the appropriate redis connection string when using `redis.password` and `redis.usePassword` (#325) (by @rebrowning)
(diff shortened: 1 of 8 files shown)


armab:
Thanks!

Based on the build, looks like Redis cluster is up & running with all the 3 instances: 
https://app.circleci.com/pipelines/github/StackStorm/stackstorm-k8s/2064/workflows/f2427912-6b96-4b76-9b69-71ccdb046f16/jobs/6554?invite=true#step-110-10

However, st2api is failing to start: https://app.circleci.com/pipelines/github/StackStorm/stackstorm-k8s/2064/workflows/f2427912-6b96-4b76-9b69-71ccdb046f16/jobs/6554?invite=true#step-109-300

Next steps, someone needs to try the same deployment and look at the st2api logs why it's failing.
Easies case if it's just a connection string issue and we need to update some templating or vars, worst case if it's about the redis compatibility with st2.

PR #319, opened 2022-07-09
Add Ability to Write/Use st2.secrets.conf from K8s Secret

cognifloyd:
This PR adds the capability to define '/etc/st2/st2.secret.conf' in a Kubernetes secret.  No longer will users need things like helm-secrets to safely store secret values that ultimately wind up in /etc/st2/*.conf.

If this is defined, the contents of key 'st2.secret.conf' within the defined secret name will be written to '/etc/st2/st2.secret.conf' on all pods.  In addition, '--config-file=/etc/st2/st2.secrets.conf' will be added to the end of the command line arguments for all pods, superseding all other configuration values.

This addresses the Comment-Based-TODO here:  https://github.com/alertlogic/stackstorm-ha/blob/master/templates/configmaps_st2-conf.yaml#L16

This replaces #289 as @ericreeves no longer has access to the @eric-al account used to create that (based on DM with Eric in slack). Plus I added unit tests.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -124,11 +124,31 @@ Reduce duplication of the st2.*.conf volume details
 - name: st2-config-vol
   mountPath: /etc/st2/st2.user.conf
   subPath: st2.user.conf
+{{- if $.Values.st2.existingConfigSecret }}
+- name: st2-config-secrets-vol
+  mountPath: /etc/st2/st2.secrets.conf
+  subPath: st2.secrets.conf
+{{- end }}
 {{- end -}}
 {{- define "stackstorm-ha.st2-config-volume" -}}
 - name: st2-config-vol
   configMap:
     name: {{ $.Release.Name }}-st2-config
+{{- if $.Values.st2.existingConfigSecret }}
+- name: st2-config-secrets-vol
+  secret:
+    secretName: {{ $.Values.st2.existingConfigSecret }}
+{{- end }}
+{{- end -}}
+
+# Override CMD CLI parameters passed to the startup of all pods to add support for /etc/st2/st2.secrets.conf
+{{- define "stackstorm-ha.st2-config-file-parameters" -}}
+- --config-file=/etc/st2/st2.conf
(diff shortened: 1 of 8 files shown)


ericreeves:
Thank you for adding the unit tests and wrapping this one up, @cognifloyd !

cognifloyd:
@armab this one is ready for review when you get a chance. Thanks!

values.yaml
```diff
@@ -84,6 +84,12 @@ st2:
   #          enabled: false
 
 
+  # Custom StackStorm config (st2.secret.conf) which will be created from the key 'st2.secret.conf' within this secret.
```
armab, line 87: Looks like the file is named `st2.secrets.conf` based on code:
```suggestion
  # Custom StackStorm config (st2.secrets.conf) which will be created from the key 'st2.secrets.conf' within this secret.
```

armab (approved):

templates/_helpers.tpl
```diff
@@ -131,11 +131,31 @@ Reduce duplication of the st2.*.conf volume details
...
+{{- end -}}
+
+# Override CMD CLI parameters passed to the startup of all pods to add support for /etc/st2/st2.secrets.conf
+{{- define "stackstorm-ha.st2-config-file-parameters" -}}
+- --config-file=/etc/st2/st2.conf
+- --config-file=/etc/st2/st2.docker.conf
+- --config-file=/etc/st2/st2.user.conf
+{{- if $.Values.st2.existingConfigSecret }}
+- --config-file=/etc/st2/st2.secrets.conf
+{{- end }}
```
cognifloyd, line 158: @Sugan-Joe, It looks like this is the place to modify the templates.
If you want to provide the file via a mount (or injected via vault, or however), then maybe we need another option. Maybe we could call it `mountedSecretsConf` (or hopefully something better)?
```suggestion
{{- if or $.Values.st2.existingConfigSecret $.Values.st2.mountedSecretsConf }}
- --config-file=/etc/st2/st2.secrets.conf
{{- end }}
```

PR #323, opened 2022-07-12
add an st2canary pod that validates st2.packs.volumes

### LINKED ISSUE #321: Create a pre-install,pre-upgrade,pre-rollback hook that validates st2.packs.volumes
The new `st2.packs.volumes` feature (introduced in #199) is extremely useful, but getting the volume config right can be confusing since Kubernetes has so many options. So, the chart should help users ensure their volume config is correct before adding/updating any other resource to kubernetes.

I think a `Pod` with a pre-* [chart hook](https://helm.sh/docs/topics/charts_hooks/) would allow the chart to validate this. Hopefully we can control the error message when the hook fails. If we can control the message, then we can explain what kinds of volumes are supported and any information we can detect about their current setup.

cognifloyd:
Closes #321

We often get questions about getting the volumes working, but the errors are very hard to debug by the time st2 is running. So, I'm trying to prevent people from installing this chart if their values for volumes do not lead to a valid setup. And, if someone tries to upgrade from an install w/o volumes enabled to an install with it enabled, then I want the upgrade to fail early if the volumes settings are not working.

There are several sources for volumes issues:

- the values are wrong
- the k8s cluster is configured incorrectly
- the underlying storage is configured incorrectly

I really want a way to push people to figure that out before the chart finishes installing something that is broken.

A key component of this is that it runs BEFORE the chart gets installed or updated. Especially with production installs, volumes issues should be found and fixed as early as possible, which is before the rest of the components get installed or upgraded.


cognifloyd:
@armab I want to get this in the next release as it should help isolate issues with `st2.packs.volumes` config and make it easier to identify and help people with that.

templates/hooks/pod_packs-volumes.yaml
```diff
@@ -0,0 +1,97 @@
...
+          rm /opt/stackstorm/packs/.write-test
+          echo
+          echo Testing write permissions on virtualenvs volume...
+          touch /opt/stackstorm/virtualenvs/.write-test
+          rm /opt/stackstorm/virtualenvs/.write-test
+          echo
+          {{- if $.Values.st2.packs.volumes.configs }}
+          echo Testing write permissions on configs volume...
+          touch /opt/stackstorm/configs/.write-test
+          rm /opt/stackstorm/configs/.write-test
```
armab, line 56: Hope that'll be helpful in debugging :+1:

armab (commented):
Considering what this Pod does, verifying write filesystem access, can we move the logic to the original `tests` dir, maybe even possible under the BATS tests?

It can be the same pod
https://github.com/StackStorm/stackstorm-k8s/blob/master/templates/tests/st2tests-pod.yaml
or perhaps a new one dedicated to volume testing?

And so it could be run with the same testing mechanism `helm test` exactly when it's needed for troubleshooting, instead of running on every Helm deployment which feels a bit off for production.

cognifloyd:
I've never thought of using `helm test` as a debugging tool. Our current test script is designed for running in CI. It would have to be a lot more robust if we want people to run it on their customized st2 clusters. For example, assuming that logging in with the st2admin user/pass is often not valid (eg w/ ldap).

mamercad:
> So, I'm trying to prevent people from installing this chart if their values for volumes do not lead to a valid setup. And, if someone tries to upgrade from an install w/o volumes enabled to an install with it enabled, then I want the upgrade to fail early if the volumes settings are not working.

I'm very much a fan of this.

mamercad:
> @armab doesn't like the idea of running this with production clusters, but that's explicitly one of my goals. Especially with production installs, volumes issues should be found and fixed as early as possible, which is before the rest of the components get installed or upgraded.

I'm not overly familiar with StackStorm HA at this point in time (I'm not running it in production yet), but, I will say that as an operator of other Kubernetes applications, I'd personally rather have an application never get off of the ground (versus being broken *while* getting off of the ground) if that's a possibility. Adding simple RW testing to a Pod whose purpose and name implies that it's for testing (heh, maybe it should be renamed to `st2canary`) seems like a fine idea to me.

In short, I'm in the "fail as early as possible" camp as well.

cognifloyd:
Ooh. I like the idea of calling it `st2canary`. Very nice. I'm going to use that.

templates/hooks/pod_packs-volumes.yaml
```diff
@@ -0,0 +1,97 @@
...
```
armab, line 56: The messaging here provides some info about what it tests, but it doesn't provide a message what to do if something fails.

As a user who has no idea about StackStorm I see the error that `/opt/stackstorm/virtualenvs/.write-test` failed, but it's hard to guess what it means next and what direction to take next.

How about adding a very loud error message saying that persistent volumes don't satisfy the write requirements and also printing a link to the docs https://github.com/StackStorm/stackstorm-k8s#method-2-shared-volumes.

values.yaml
```diff
@@ -135,6 +135,11 @@ st2:
     # The choice of storage solution is cluster-dependent (it changes besed on where the cluster is hosted
     # and which storage solutions are available in your cluster).
     #
+    # WARNING: These volumes must use ReadWriteMany (RWX) access mode so StackStorm
+    # can dynamically install packs from any of the st2actionrunner pods and have those
+    # file changes available in all of the other StackStorm pods.
+    # see: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes
```
armab, line 141: This is a good message example to include in the logs too :+1: in case of st2canary failed

values.yaml
```diff
@@ -962,6 +967,22 @@ jobs:
     #     - run
     #     - --tail
     #     - custom_pack.warn_about_upgrade
+
+# st2canary has settings used for temporary chart hook resources
+# like a pod that validates packs.volumes before installs/upgrades.
```
armab, line 972: Just a bit more info about the `st2canary` would be good here and mentioning a fact the pod will be deleted if the configuration is good.

armab (changes requested):
Left a few more comments.

Can we make it a `Job`, same as others that we already have and call more explicitly what it does like `st2-verify-pack-write-access`?

The jobs we already have: 
https://github.com/StackStorm/stackstorm-k8s/blob/master/templates/jobs.yaml
and as I remember if a job like register-content failed, the entire helm deployment will report a failure too. So logically that's following what you want to do.


Screenshot 2022-12-21 at 21-22-31 add an st2canary pod that validates st2 packs volumes · StackStorm_stackstorm-k8s@21a961c

That would be a bit more consistent with the existing mechanics.

cognifloyd:
Yeah, that makes sense. I'll work on:
- turning it into a job
- adding more detailed error messaging with:
    - a link to the docs
    - what to do in case of failure

values.yaml
cognifloyd: Included.

templates/hooks/pod_packs-volumes.yaml
cognifloyd: OK. I updated the script to hopefully be both clearer and more helpful.

values.yaml
```diff
@@ -962,6 +967,22 @@ jobs:
...
+st2canary:
```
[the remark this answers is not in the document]
cognifloyd: k. dropped.

templates/jobs.yaml
```diff
@@ -484,6 +484,125 @@ spec:
...
+        {{- if $.Values.st2.packs.volumes.configs }}
+          {{/* do not include the pack-configs-volume helper here */}}
+        - name: st2-pack-configs-vol
+          {{- toYaml $.Values.st2.packs.volumes.configs | nindent 10 }}
+        {{- end }}
+        {{- range $.Values.st2canary.extra_volumes }}
+        - name: {{ required "Each volume must have a 'name' in st2canary.extra_volumes" .name }}
+          {{- tpl (required "Each volume must have a 'volume' definition in st2canary.extra_volumes" .volume | toYaml) $ | nindent 10 }}
+        {{- end }}
+    {{- if $.Values.dnsPolicy }}
```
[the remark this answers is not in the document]
cognifloyd: k. dropped

templates/jobs.yaml
```diff
@@ -484,6 +484,125 @@ spec:
     {{- end }}
 
 {{- end }}
+{{- if $.Values.st2.packs.volumes.enabled }}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ $.Release.Name }}-job-st2canary-for-writable-packs-volumes
```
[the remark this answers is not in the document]
cognifloyd: renamed to `-job-ensure-packs-volumes-are-writable`

PR #324, opened 2022-07-16
Address issue with register content when configuring volumes

### LINKED ISSUE #320: Problem with pvc volume mount on job-st2-register-content
I am new to stackstrom trying to use volume mount in order to enable write option using pvc.

Always get job-st2-register-content keeps on waiting to start.
  I referred issue 238 which was having similar issue but with NFS deploy.

Any help is much appreciated. Thanks!!

rebrowning:
add the additional check to the condition, since what is happening is the container is attempting to run but the st2-pack-configs-from-helm-vol is not being created because the condition is not met

I've been digging in to try and understand what the init containers are doing, but the root of the problem is the init container is expecting a volume that doesn't exist as part of the pod as part of a condition, so I simply added the condition to the init container causing the issue.

If there's a better way to resolve it please let me know and I'll take a look.

Fixes #320


cognifloyd:
Wow! Good find. Yeah, that's clearly a bug.

cognifloyd (approved):

cognifloyd:
So, we have unit tests for a lot of the chart, but not for the init containers. `helm-unittest` requires an index instead of a container name which makes it far too brittle to write unit tests. So, the only way to identify init container bugs is manually. Thank you for finding this! I'll merge it once tests pass.

PR #325, opened 2022-08-03
fix for the redis password not being set in the connection string

rebrowning:
I'd like some feedback on how we want to handle this.

There are other options with how to configure redis in the underlying bitnami chart (generated password, secretfile). However I don't know how to grab either the generated helm value or the value from the secret file to then inject it into the redis url. Are there some additional coordination parameters we can use in st2 to set the password or an environment variable that st2 will consume for the redis password?

Additionally, you can set sentinel to be protected by the same password as well, however this `DOES NOT` work due to the tooz driver not populating the password for the sentinel client, only for the redis client that gets created from the sentinel get master request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fix mounts for `jobs.preRegisterContentCommand` container to use the same mounts as the primary register-content container. (#322) (by @cognifloyd)
 * Add support for providing custom st2actionrunner-specific docker repository, image name, pull policy, and pull secret via `values.yaml`. (#141) (by @Sheshagiri)
 * Fix bug that hung an init container when `st2.packs.volumes.enabled` without `st2.packs.volumes.configs`. (#324) (by @rebrowning)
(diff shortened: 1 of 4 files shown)


cognifloyd (approved):
I don't use the subchart, so I can't say which method will be better to use here. From reading, this looks good to me, but I'd prefer someone else who uses the redis subchart chime in before I merge this.

arms11:
Hi @rebrowning I am one of the users who uses the sub-chart. As long as you have tested the deployment without password and it works, I think it's a good change. We want to make sure it's backward compatible. Could you please confirm?

robertorch:
@arms11 I tested both with and without a password configuration for Redis when doing this. I'm more than happy to provide a couple screenshots after hours of the config/template/deployment success for completeness.

arms11:
Nope. I appreciate the response.

cognifloyd:
Please also add a changelog entry.

Issue #326, opened 2022-08-10
Add ability to create custom labels for service account


SuganJoe:
We have a requirement internally in our organization to add custom labels to the service account being created by st2. [Here](https://github.com/StackStorm/stackstorm-k8s/blob/master/templates/service-account.yaml#L11). 

Currently, it seems like there is no way to pass custom label values  and it only accepts templating out a list of stack storm service names, please correct me if I'm wrong.

I would like to contribute to add the ability to accept custom values here.

cognifloyd:
Nothing in this chart accepts custom labels. We do however support custom annotations.

I'm very curious about your use case. Would annotations work? Or are you dealing with a corporate policy that requires labels? Or something else?

SuganJoe:
Internally we have a requirement to have a custom label to the service account (SA) in order to get the SA or the pod associated with SA access to internal vault.

In other words, when a SA has a specific label, it would help the corresponding pod to obtain access to vault storage.

cognifloyd:
Cool. I'm open to merging a PR if you can work on it.

You need to adjust the templates, document the relevant values in values.yaml, and add unit tests for it similar to:
https://github.com/StackStorm/stackstorm-k8s/blob/master/tests/unit/custom_annotations_test.yaml

You can see the current labels tests in this file:
https://github.com/StackStorm/stackstorm-k8s/blob/master/tests/unit/labels_test.yaml

SuganJoe:
yes, will work on it. you can assign the issue to me.

PR #327, opened 2022-08-11
Adding the ability to service account to accept custom labels

SuganJoe:
This PR is about adding the ability to service-account.yaml template to accept custom labels defined through values.yaml

Closes: #326

Use case:
Internally we have a requirement to have a custom label to the service account (SA) in order to get the SA or the pod associated with SA access to internal vault. In other words, when a SA has a specific label, it would help the corresponding pod to obtain access to vault storage.

### DIFF
--- a/tests/unit/labels_test.yaml
+++ b/tests/unit/labels_test.yaml
@@ -219,10 +219,10 @@ tests:
     set:
       serviceAccount:
         create: true
-        serviceAccountLabels: &labels
+        serviceAccountLabels:
           foo: bar
           answer: "42"
-    asserts: &labels_asserts
+    asserts:
       - isNotNull:
           path: metadata.labels
       - equal:
--- a/tests/unit/labels_test.yaml
+++ b/tests/unit/labels_test.yaml
@@ -214,6 +214,24 @@ tests:
           path: metadata.labels.heritage
           value: Helm
 
+  - it: ServiceAccount accepts custom labels
+    template: service-account.yaml
+    set:
+      serviceAccount:
+        create: true
+        serviceAccountLabels: &labels
+          foo: bar
+          answer: "42"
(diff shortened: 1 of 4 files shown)


tests/unit/labels_test.yaml
```diff
@@ -214,6 +214,24 @@ tests:
...
 
+  - it: ServiceAccount accepts custom labels
+    template: service-account.yaml
+    set:
+      serviceAccount:
+        create: true
+        serviceAccountLabels: &labels
+          foo: bar
+          answer: "42"
+    asserts: &labels_asserts
```
cognifloyd, line 225:
```suggestion
    asserts:
```

tests/unit/labels_test.yaml
```diff
@@ -214,6 +214,24 @@ tests:
           path: metadata.labels.heritage
           value: Helm
 
+  - it: ServiceAccount accepts custom labels
+    template: service-account.yaml
+    set:
+      serviceAccount:
+        create: true
+        serviceAccountLabels: &labels
```
cognifloyd, line 222:
```suggestion
        serviceAccountLabels:
```

cognifloyd (commented):
You only need the `&labels...` yaml anchors when reusing bits with `*labels...` so drop that. At a glance I don't see any other issues.

tests/unit/labels_test.yaml
SuganJoe: 178c74e63f151f5304999970e8ad70a4a84da9e7

tests/unit/labels_test.yaml
SuganJoe: 178c74e63f151f5304999970e8ad70a4a84da9e7

cognifloyd (approved):

PR #328, opened 2022-09-05
Clarify docs around LDAP group membership and RBAC roles

anrajme:
Clarity to the documentation for automatically granting Roles Based on LDAP Group Membership

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -258,6 +258,7 @@ st2:
       #  description: "Automatically grant admin role to all stormers group members."
       #  roles:
       #    - "admin"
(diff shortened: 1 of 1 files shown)


cognifloyd:
I don't think that's the best place for that bit of documentation. Maybe we can add a link to the docs? Maybe a warning to LDAP users is in order, but where. I'm not sure.

ribal-aladeeb:
I would suggest having a one-liner warning about this in the README to start and then a possible link to a longer-winded explanation in the docs

cognifloyd:
@anrajme Are you planning to do any more on this PR?

cognifloyd:
No feedback in a while. I'm closing this. I'd you want we can reopen and discuss more.

PR #333, opened 2022-10-12
Enable Extra Volumes in Jobs

bmarick:
My database connection required me to provide a certificate trust chain to validate the connection.  \
I am able to set that certificate in the deployments by using the `extra_volumes` options in the values file. \
However, the Jobs do not have that option currently available.

This is my attempt to enable the extra_volume support to Jobs.
I have tested the changes in my cluster and validated that the volume mounts worked for my use case.

### DIFF
--- a/tests/unit/jobs_test.yaml
+++ b/tests/unit/jobs_test.yaml
@@ -1,0 +1,47 @@
+---
+suite: Custom Annotations
+templates:
+  - jobs.yaml
+
+  # included templates must also be listed
+  - configmaps_overrides.yaml
+  - configmaps_packs.yaml
+  - configmaps_rbac.yaml
+  - configmaps_st2-conf.yaml
+  - configmaps_st2-urls.yaml
+  - configmaps_st2web.yaml
+  - secrets_datastore_crypto_key.yaml
+  - secrets_ssh.yaml
+  - secrets_st2apikeys.yaml
+  - secrets_st2auth.yaml
+  - secrets_st2chatops.yaml
+
+tests:
+  - it: Jobs add extra_volumes
+    template: jobs.yaml
+    set:
+      jobs:
+        extra_volumes:
+          - name: custom-config-map-volume
+            mount:
+              mountPath: /config/config.yml
+              subPath: config.yml
+            volume:
+              configMap:
+                name: customConfig
+    asserts:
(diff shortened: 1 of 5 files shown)


armab:
Good stuff, +1 for the feature.

Looks like the tests are failing though and need an update too.

bmarick:
@armab  Unit Tests have been fixed can you merge this change?

tests/unit/jobs_test.yaml
```diff
@@ -0,0 +1,47 @@
+---
+suite: Custom Annotations
```
copart-jafloyd, line 2:
```suggestion
suite: Extra Volumes
```

Could we rename this file to `extra_volumes_test.yaml`?

Each file tests one feature across all applicable templates. We can add more extra_volumes tests in another PR for all the deployments that use this too.

values.yaml
```diff
@@ -919,6 +919,10 @@ jobs:
   # HTTP_PROXY: http://proxy:1234
   ## These named secrets (managed outside this chart) will be added to envFrom.
   envFromSecrets: []
+  # mount extra volumes on the st2web pod(s) (primarily useful for k8s-provisioned secrets)
```
copart-jafloyd, line 922:
```suggestion
  # mount extra volumes on the jobs pods (primarily useful for k8s-provisioned secrets)
```

copart-jafloyd (changes requested):
This is almost ready to merge - just a few minor suggestions, and then it looks good!

cognifloyd:
Ooops. @copart-jafloyd is me. I reviewed in the wrong window. Sorry!

armab:
Thanks for the reviews @cognifloyd & @copart-jafloyd ! :smile_cat:

tests/unit/jobs_test.yaml
bmarick: Both file and suite name have been updated

values.yaml
bmarick: Comment should be updated

bmarick:
@cognifloyd & @copart-jafloyd  the changes you have requested have been completed and the tests are passing. 
 Are there any additional changes you would like to see?

cognifloyd (approved):
LGTM (this time reviewing with the right account 😉)

PR #341, opened 2022-12-19
Add K3s End-to-End Workflow in GitHub (Experimental)

### LINKED ISSUE #243: CI/CD: Switch to lightweight Kubernetes distribution
Currently, we're relying on CircleCI large container size that provides 4CPUs & 15GB RAM ~which is a paid feature~ (Update: it's free now):
https://github.com/StackStorm/stackstorm-ha/blob/3ba85e6c584229e21cf41e2d19ec14f2bef4ffcc/.circleci/config.yml#L52-L55

1) Switch to a smaller CircleCI container size
2) Replace K8s with a lightweight distribution like K3s or K0s that'll fit CI/CD environment better:
- https://k3s.io/
- https://github.com/k0sproject/k0s/
- https://kind.sigs.k8s.io/

~Hopefully, this could also help with the issues when builds reported random stability failures caused by etcd/K8s/CircleCI~ (Update: it runs well recently).

mamercad:
Add an experimental workflow in GitHub using K3s for testing; fixes #243. Here's an example in my [fork](https://github.com/mamercad/stackstorm-k8s/actions/runs/3731588613/jobs/6330007930).

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -1,0 +1,115 @@
+name: helm
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+  push:
+    branches:
+      - master
+  schedule:
+    - cron: "0 1 * * *"
+  workflow_dispatch:
+
+jobs:
+  helm-lint:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 10
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+
+      - name: Update stackstorm-ha chart dependencies
+        run: |
+          set -x
+          helm dependency update
+
+      - name: Helm lint
+        run: |
+          helm lint
+
+      - name: Cache community
+        id: cache-community
+        uses: actions/cache@v3
+        with:
+          path: community
+          key: ${{ runner.os }}-community-${{ hashFiles('conf/**', 'templates/**', 'Chart.yaml', 'values.yaml') }}
+
+      - name: Helm template
+        if: steps.cache-community.outputs.cache-hit != 'true'
+        shell: bash
+        run: |
+          helm template --output-dir community .
+
+  k8s-lint:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 10
+    needs: [helm-lint]
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+
+      - name: Cache community
+        id: cache-community
+        uses: actions/cache@v3
+        with:
+          path: community
+          key: ${{ runner.os }}-community-${{ hashFiles('conf/**', 'templates/**', 'Chart.yaml', 'values.yaml') }}
+
+      - name: Kubernetes kubeval lint
+        uses: instrumenta/kubeval-action@master
+        with:
+          files: community
+
+  helm-e2e:
+    needs: [k8s-lint]
+    runs-on: ubuntu-22.04
+    # NOTE: Just a thought in case the timeouts fail; might not be
+    # necessary, but might not hurt either, would vary based on the
+    # size of the testing matrix, too.
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      max-parallel: 1
+      matrix:
+        # TODO: Document which versions we support and cover them.
+        k3s-channel:
+          - "v1.25.4+k3s1"
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+
+      - name: Set up K3s
+        id: k3s
+        uses: jupyterhub/action-k3s-helm@v3
+        with:
+          k3s-channel: ${{ matrix.k3s-channel }}
+
+      - name: Update stackstorm-ha chart dependencies
+        run: |
+          set -x
+          helm dependency update
+
+      - name: Helm install
+        run: |
+          helm install --timeout 10m0s --debug --wait \
+            --name-template stackstorm-ha .
+
+      - name: Helm test
+        run: |
+          helm test stackstorm-ha
+
+      - name: Helm upgrade with RBAC enabled
+        run: |
+          helm upgrade --set st2.rbac.enabled=true \
+            --timeout 5m0s --debug --wait stackstorm-ha .
+
+      - name: Helm test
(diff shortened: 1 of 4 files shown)


.github/workflows/helm.yaml
```diff
@@ -0,0 +1,115 @@
+name: helm
```
cognifloyd, line 1: This needs a better name.

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,115 @@
...
+        with:
+          path: community
+          key: ${{ runner.os }}-community-${{ hashFiles('conf/**', 'templates/**', 'Chart.yaml', 'values.yaml') }}
+
+      - name: Kubernetes kubeval lint
+        uses: instrumenta/kubeval-action@master
+        with:
+          files: community
+
+  helm-e2e:
```
cognifloyd, line 65: I'm concerned that this job to 15 min. In CircleCI it takes around 6-7 minutes. Do you know what makes the difference?

cognifloyd (commented):
Exciting stuff

.github/workflows/helm.yaml
mamercad: > This needs a better name.

Hehe, yes it does 😄

cognifloyd: Maybe "Lint & E2E Tests"?

mamercad:
> BTW, speaking of CircleCI Issue, - #243 it's a bit outdated (sorry for that!). CircleCI is free for us now and is using `'large' 4 vCPUs & 15GB RAM` executor. Added more context here: [#243 (comment)](https://github.com/StackStorm/stackstorm-k8s/issues/243#issuecomment-1357962293)
> 
> We're fine running `minikube` w CircleCI. Consistent runs for some good time and that environment probably provides the best testing compatibility experience with the real K8s cluster.
> 
> But perhaps adding `K3s` as an experimental run would also be nice for the community? The flip side is that it may add some support burden if something fails in k3s?

Sweet, glad CircleCI is `$free`. As far as the support burden goes, we can label the workflow as "experimental" and not require it to pass, right?

.github/workflows/helm.yaml
mamercad: Hrm, the [most recent run](https://github.com/StackStorm/stackstorm-k8s/actions/runs/3735681553/jobs/6339213088) was just a handful of seconds.

mamercad:
> Let me update the changelog and readme ...

Should be all set, sorry for not catching this @cognifloyd.

mamercad:
We should probably bug this (unless it's already covered somewhere and I missed it):

```
# TODO: Document which versions we support and cover them.
```

cognifloyd:
> We should probably bug this (unless it's already covered somewhere and I missed it):
> 
> ```
> # TODO: Document which versions we support and cover them.
> ```

Yes, please create an issue for it. I'm not sure what range of k8s versions we support.

armab:
Looking at Screenshot 2022-12-21 at 00-09-02 StackStorm_stackstorm-k8s K8s Helm Chart that codifies StackStorm (aka IFTTT for Ops https __stackstorm com_) Highly Availability fleet as a simple to use reproducible infrastructure-as-code app


`k3s` is what really unique about this pipeline and distinguishes it from the existing CircleCI e2e tests :+1: 
We probably don't need to run the linting in both CircleCI and Github.


Thinking that it would be nice to have a consistent naming across the checks and show more specifics about the environment:

- ci / helm-unittest
- ci / helm-lint
- ci / k8s-lint 
- ci / helm-e2e (v1.25.4+k3s1)
- ci/circleci: helm-e2e (vX.Y.Z+minikube)
- license/cla

mamercad:
> Looking at the result Screenshot 2022-12-21 at 00-09-02 StackStorm_stackstorm-k8s K8s Helm Chart that codifies StackStorm (aka IFTTT for Ops https _stackstorm com) Highly Availability fleet as a simple to use reproducible infrastructure-as-code app
> 
> `k3s` is what really unique about this pipeline and distinguishes it from the existing CircleCI e2e tests 👍 We probably don't need to run the linting in both CircleCI and Github.
> 
> Thinking that it would be nice to have a consistent naming across the checks and show more specifics about the environment, leaving only `e2e` minikube on CircleCI:
> 
> * ci / helm-unittest
> * ci / helm-lint
> * ci / k8s-lint
> * ci / helm-e2e (v1.25.4+k3s1)
> * ci/circleci: helm-e2e (vX.Y.Z+minikube)
> * license/cla

I'm on board, these are good suggestions for the next iteration!

cognifloyd:
For GHA, the word before the first `/` is the name of the workflow.

For external reporters, the service reports its name before the `:`. CircleCI chose to report their name with the `ci/` prefix.

I think listing "CI" is redundant. GHA is CI. All the listed checks are CI. Ideally I would like to see something meaningful for the names of workflows and jobs. We're stuck with the pointless `ci/` prefix on `ci/circleci`. But everything else can be more helpful.

Currently we have lint combined with e2e tests to avoid running the e2e tests unless the lint checks pass. If we are willing to skip that optimization, then I propose we split things into these workflows and jobs:

- GHA
    - Lint workflow
        - helm-lint job
        - k8s-lint job
    - Unit tests workflow
        - helm-unittest job
    - E2E Tests workflow
        - k3s job
- CircleCI
    - E2E Tests minikube

That would lead to something that looks like this:

- Lint / helm-lint (pull_request)
- Lint / k8s-lint (pull_request)
- Unit Tests / helm-unittest (pull_request)
- E2E Tests / k3s (experimental) (v1.25.4+k3s1) (pull_request)
- ci/circleci: E2E Tests / minikube

For CircleCI, we can add the slash in our workflow name, assuming they don't disallow that character, to make it read more like the pattern GHA supplies.

If, on the other hand, you want the `CI / ` prefix, then we have to either:
1. prefix the workflow names with CI (making it `CI / Lint` and `CI / E2E Tests`), or
2. combine all our GHA workflows into a single workflow named "CI".

I do not want to combine the workflows (option 2) because that makes restarting them more painful, and it makes the workflow file much longer and harder to maintain.

PR #344, opened 2022-12-21
CI: Reorganize and rename GH and Circle workflows

### LINKED ISSUE #343: Reorganize CI
For GHA, the word before the first `/` is the name of the workflow.

For external reporters, the service reports its name before the `:`. CircleCI chose to report their name with the `ci/` prefix.

I think listing "CI" is redundant. GHA is CI. All the listed checks are CI. Ideally I would like to see something meaningful for the names of workflows and jobs. We're stuck with the pointless `ci/` prefix on `ci/circleci`. But everything else can be more helpful.

Currently we have lint combined with e2e tests to avoid running the e2e tests unless the lint checks pass. If we are willing to skip that optimization, then I propose we split things into these workflows and jobs:

- GHA
    - Lint workflow
        - helm-lint job
        - k8s-lint job
    - Unit tests workflow
        - helm-unittest job
    - E2E Tests workflow
        - k3s job
- CircleCI
    - E2E Tests minikube

That would lead to something that looks like this:

- Lint / helm-lint (pull_request)
- Lint / k8s-lint (pull_request)
- Unit Tests / helm-unittest (pull_request)
- E2E Tests / k3s (experimental) (v1.25.4+k3s1) (pull_request)
- ci/circleci: E2E Tests / minikube

For CircleCI, we can add the slash in our workflow name, assuming they don't disallow that character, to make it read more like the pattern GHA supplies.

If, on the other hand, you want the `CI / ` prefix, then we have to either:
1. prefix the workflow names with CI (making it `CI / Lint` and `CI / E2E Tests`), or
2. combine all our GHA workflows into a single workflow named "CI".

I do not want to combine the workflows (option 2) because that makes restarting them more painful, and it makes the workflow file much longer and harder to maintain.

_Originally posted by @cognifloyd in https://github.com/StackStorm/stackstorm-k8s/issues/341#issuecomment-1360639154_

mamercad:
Striving for:

* Lint / helm-lint (pull_request)
* Lint / k8s-lint (pull_request)
* Unit Tests / helm-unittest (pull_request)
* E2E Tests / k3s (experimental) (v1.25.4+k3s1) (pull_request)
* ci/circleci: E2E Tests / minikube

For Circle, the best we can do (because of how the integration works, and, naming restrictions) is likely "ci/circleci: e2e-minikube", mentioned [here](https://github.com/StackStorm/stackstorm-k8s/issues/343#issuecomment-1361345280).

Fixes #343.


cognifloyd (approved):
Minor not. Otherwise LGTM

cognifloyd (approved):

cognifloyd:
K. I adjusted the branch protection to use the new job/workflow names.
I did not add the k3s job as required yet. Once we're ready to say it's not experimental, we can require it as well.

mamercad:
> 

Excellent, sounds good!

mamercad:
Ahhh, I think we need to update the Badges, stand by.

Issue #342, opened 2022-12-20
Document and cover supported Kubernetes versions


mamercad:
> ```
> # TODO: Document which versions we support and cover them.
> ```

Yes, please create an issue for it. I'm not sure what range of k8s versions we support.

_Originally posted by @cognifloyd in https://github.com/StackStorm/stackstorm-k8s/issues/341#issuecomment-1358827972_

mamercad:
Kubernetes release information is [here](https://kubernetes.io/releases/).

At the time of this comment:
> The Kubernetes project maintains release branches for the most recent three minor releases (1.26, 1.25, 1.24).

mamercad:
Closed by #346.

PR #346, opened 2022-12-23
CI: Cover the three most recent Kubernetes versions in Minikube and the single most recent in K3s

mamercad:
Relates to #342.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,9 @@ orbs:
 jobs:
   # Spin up minikube K8s cluster and run Helm chart & e2e tests on it
   e2e-minikube:
+    parameters:
+      minikube-version:
+        type: string
     # 'large' 4 vCPUs & 15GB RAM CircleCI machine executor
     # required to deploy heavy 'stackstorm-ha' Helm release with RabbitMQ, MongoDB, Redis clusters and 25+ st2 Pods.
     # https://circleci.com/docs/2.0/configuration-reference/#machine-executor-linux
@@ -24,7 +27,7 @@ jobs:
       - kubernetes/install
       - minikube/minikube-install:
           # https://github.com/kubernetes/minikube/releases
-          version: v1.15.1
+          version: << parameters.minikube-version >>
       - run:
           name: Install dependencies
           command: |
@@ -64,11 +67,23 @@ workflows:
   version: 2
   e2e:
     jobs:
-      - e2e-minikube
+      - e2e-minikube:
+          matrix:
+            parameters:
+              minikube-version: # https://github.com/kubernetes/minikube/releases
+                - v1.22.0
+                - v1.21.0
+                - v1.20.0
   # Run periodic nightly Helm tests to ensure there are no regressions
   e2e-nightly:
     jobs:
-      - e2e-minikube
+      - e2e-minikube:
+          matrix:
+            parameters:
+              minikube-version: # https://github.com/kubernetes/minikube/releases
+                - v1.22.0
+                - v1.21.0
+                - v1.20.0
     triggers:
       - schedule:
           cron: "0 1 * * *"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,8 +13,6 @@ jobs:
   # Spin up minikube K8s cluster and run Helm chart & e2e tests on it
   e2e-minikube:
     parameters:
-      minikube-version:
-        type: string
       kubernetes-version:
         type: string
     # 'large' 4 vCPUs & 15GB RAM CircleCI machine executor
@@ -29,7 +27,7 @@ jobs:
       - kubernetes/install
       - minikube/minikube-install:
           # https://github.com/kubernetes/minikube/releases
-          version: << parameters.minikube-version >>
+          version: v1.28.0
       - run:
           name: Install Helm v3
           command: curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash
@@ -41,7 +39,7 @@ jobs:
           command: helm dependency update
       - run:
           name: Helm install stackstorm-ha chart
-          command: helm install --timeout 10m0s --debug --wait --name-template stackstorm-ha .
+          command: helm install --timeout 5m0s --debug --wait --name-template stackstorm-ha .
       - run:
           name: Helm test
           command: helm test stackstorm-ha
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e.yaml
```diff
@@ -26,8 +26,16 @@ jobs:
...
+        # https://github.com/k3s-io/k3s/releases
         k3s-channel:
+          # - "v1.18.13+k3s1"
+          # - "v1.19.5+k3s1"
+          # - "v1.20.2+k3s1"
+          # - "v1.21.0+k3s1"
+          # - "v1.23.6+k3s1"
+          - "v1.24.4+k3s1"
           - "v1.25.4+k3s1"
+          - "v1.26.0+k3s1"
```
armab, line 38: I think having a K8s compatibility matrix would be more suitable for something like minikube as that better mimics the real K8s cluster and k3s is more customized and we may want just to check if the latest k3s works.

mamercad: I don't quite follow the argument here ... we're not proving whether or not k3s works ... and I don't know what you mean by "better mimics a real K8s cluster". According to [this](https://docs.k3s.io/) "K3s is a fully compliant Kubernetes distribution".

mamercad: The point of this matrix is to verify that `stackstorm-k8s` works under Kubernetes 1.24, 1.25, and 1.26.

armab: k3s is a modified lightweight 3rd-party distribution of K8s.
While it should be fully compatible, the reality there are always unique issues someone can hit running K3s compared to running an official K8s cluster. See: https://github.com/k3s-io/k3s/issues

`minikube` is re-using `kubeadm` and other primitives under the hood and is more comparable to running an official k8s cluster.

StackStorm is not an IoT platform and ideally if we test against environments closer to its real-world usage. k3s is ok for experimental compatibility check, but I'd prefer if we run the K8s build matrix with `minikube`.

Here is the ideal build matrix: 
- minikube-k8s-v1.24
- minikube-k8s-v1.25
- minikube-k8s-latest
- k3s-experimental-latest

Hope that gives more context ;)

mamercad: We'll just have to agree to disagree, `k3s` is [certified to be conformant](https://www.cncf.io/certification/software-conformance/).

armab: There is always a diff and unique issues that are present in K3s, but not in the K8s and vice-versa.

If something fails in tests in K3s that does work with `minikube` or prod K8s cluster I'd like to avoid the rabbit hole and support burden for us. And vice-versa, if tests passed with the k3s distro and the user comes up with the issue that something didn't work in the official K8s, - I don't want it for maintainers debugging this.

I'd like to be closer in tests to the real K8s environment where StackStorm is most likely to be deployed, and we can do it already with minikube that we're already running.

armab: BTW we tried KinD (another conformant distro), had some issues in the past, but it might work better now.

Do you have specific requirements to run and test StackStorm under the `k3s` distro, something that's required by your environment?

Or would you just like having different supported K8s versions in the build matrix?

mamercad: Personally, I do run `k3s` in production (and, I know others that do as well). I understand where you are coming from, and I'm fine with dropping `k3s` if that's what you want, but I'd kindly ask that you not slight their project by making statements that it's "not ready for production" or potentially only suitable for "edge or IoT". If you could spare the time, I think [this](https://traefik.io/glossary/k3s-explained/) is worthy of a quick read.

mamercad: When you said "kubeadm and other primitives under the hood and is more comparable to running an official k8s cluster" what I was thinking was "we're not testing the mechanism by which Kubernetes is stood up; what we should be testing is that stackstorm-ha successfully deploys to Kubernetes" -- and of course, the current test is doing that. In my mind, since the CNCF has officially adopted `k3s` as a distribution, I'll take that at face-value that it's (`k3s`) is a viable and *supported* distribution. Again, I personally don't care if we test against `k3s` or `minikube`, or both.

mamercad: When you mentioned to "See: https://github.com/k3s-io/k3s/issues", what'd I'd ask is, which of those issues specifically pertains to "incompatibility with upstream proper"?

.circleci/config.yml
```diff
@@ -64,11 +67,23 @@ workflows:
...
   e2e:
     jobs:
-      - e2e-minikube
+      - e2e-minikube:
+          matrix:
+            parameters:
+              minikube-version: # https://github.com/kubernetes/minikube/releases
+                - v1.22.0
+                - v1.21.0
+                - v1.20.0
```
armab, line 76: Changelog says we're testing with 3 the latest K8s versions, while the matrix is about minikube versions which is not the same.

I saw you debugged the latest `minikube` (and all the cri-o stuff) as it doesn't work anymore with the default engine as I understand.

To assist, the docker driver looks to be working with the latest minikube with no customizations: 
https://github.com/StackStorm/stackstorm-k8s/compare/test-minikube?expand=1#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47

We can use that to pin the latest minikube, and add 3 K8s versions to the matrix then as originally thought.

armab (changes requested):
Thanks for working on that and all the debugging!

Left one change comment to assist, so we can really test with the 3 K8s versions and just the latest minikube (can even unpin it, if you'd like).

.circleci/config.yml
mamercad: Yep, thanks for the help on that, I'm iterating...

mamercad: I'm not sure why one of the checks isn't returning status, but, I think we're passing now.

mamercad: > We can use that to pin the latest minikube, and add 3 K8s versions to the matrix then as originally thought.

Ah ... I'll do this next.

.circleci/config.yml
```diff
@@ -64,11 +58,25 @@ workflows:
   version: 2
   e2e:
     jobs:
-      - e2e-minikube
+      - e2e-k8s:
+          matrix:
+            parameters:
+              # https://kubernetes.io/releases
+              kubernetes-version:
+                - "v1.26.0"
```
armab, line 66: Looks like it's [failing](https://app.circleci.com/pipelines/github/mamercad/stackstorm-k8s/9/workflows/9e65320a-bfcf-45f1-b6f4-730bf596e5c9/jobs/16) on K8s v1.26.0: 
```
! Specified Kubernetes version 1.26.0 is newer than the newest supported version: v1.25.3.
Use `minikube config defaults kubernetes-version` for details.
```
It was just released a month ago.

Let's perhaps swap the version?
```suggestion
                - "v1.23.15"
```

armab (approved):
Looks great!

Just one comment about the K8s version.

.circleci/config.yml
mamercad: Sounds good, pushed!

PR #356, opened 2023-02-01
configurable utilityImage + clusterDomain and increase default db timeouts to avoid replicaset timeout

guzzijones:
allow alternate proxy registry to be used with helm chart.
fix replica set timeout error
fix dns resolution errors for mongo and redis

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -103,7 +103,7 @@ Generate comma-separated list of nodes for MongoDB-HA connection string, based o
 {{- range $index0 := until $replicas -}}
   {{- $index1 := $index0 | add1 -}}
   {{- if eq $architecture "replicaset" }}
-    {{- $mongo_fullname }}-{{ $index0 }}.{{ $mongo_fullname }}-headless.{{ $.Release.Namespace }}.svc.cluster.local{{ if ne $index1 $replicas }},{{ end }}
+    {{- $mongo_fullname }}-{{ $index0 }}.{{ $mongo_fullname }}-headless.{{ $.Release.Namespace }}.svc.{{ $.Values.clusterDomain }}{{ if ne $index1 $replicas }},{{ end }}
   {{- else }}
     {{- $mongo_fullname }}-{{ $index0 }}.{{ $mongo_fullname }}{{ if ne $index1 $replicas }},{{ end }}
   {{- end -}}
@@ -122,9 +122,9 @@ Generate list of nodes for Redis with Sentinel connection string, based on numbe
 {{- $sentinel_port := (index .Values "redis" "sentinel" "port") }}
 {{- range $index0 := until $replicas -}}
   {{- if eq $index0 0 -}}
-    {{ $.Release.Name }}-redis-node-{{ $index0 }}.{{ $.Release.Name }}-redis-headless.{{ $.Release.Namespace }}.svc.cluster.local:{{ $sentinel_port }}?sentinel={{ $master_name }}
+    {{ $.Release.Name }}-redis-node-{{ $index0 }}.{{ $.Release.Name }}-redis-headless.{{ $.Release.Namespace }}.svc.{{ $.Values.clusterDomain }}:{{ $sentinel_port }}?sentinel={{ $master_name }}
   {{- else -}}
-    &sentinel_fallback={{ $.Release.Name }}-redis-node-{{ $index0 }}.{{ $.Release.Name }}-redis-headless.{{ $.Release.Namespace }}.svc.cluster.local:{{ $sentinel_port }}
+    &sentinel_fallback={{ $.Release.Name }}-redis-node-{{ $index0 }}.{{ $.Release.Name }}-redis-headless.{{ $.Release.Namespace }}.svc.{{ $.Values.clusterDomain }}:{{ $sentinel_port }}
   {{- end -}}
 {{- end -}}
 {{- end -}}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -46,7 +46,7 @@ Generate Docker utility image line
 {{- if .Values.image.utilityImage -}}
 {{ .Values.image.utilityImage }}
 {{- else -}}
-'docker.io/library/busybox:1.28'
+docker.io/library/busybox:1.28
 {{- end -}}
 {{- end -}}
 
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -40,6 +40,18 @@ app.kubernetes.io/instance: {{ $root.Release.Name }}
 {{- end -}}
 
 {{/*
+Generate Docker image registry: container registry
+*/}}
+{{- define "stackstorm-ha.imageRegistry" -}}
+{{- if .Values.image.registry -}}
+{{ .Values.image.registry }}
+{{- else -}}
+'docker.io'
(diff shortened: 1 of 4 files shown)


guzzijones:
Not sure why the key is missing for the checks

templates/_helpers.tpl
```diff
@@ -91,7 +103,7 @@ Generate comma-separated list of nodes for MongoDB-HA connection string, based o
 {{- range $index0 := until $replicas -}}
   {{- $index1 := $index0 | add1 -}}
   {{- if eq $architecture "replicaset" }}
-    {{- $mongo_fullname }}-{{ $index0 }}.{{ $mongo_fullname }}-headless{{ if ne $index1 $replicas }},{{ end }}
+    {{- $mongo_fullname }}-{{ $index0 }}.{{ $mongo_fullname }}-headless.{{ $.Release.Namespace }}.svc.cluster.local{{ if ne $index1 $replicas }},{{ end }}
```
cognifloyd, line 106: This would break my setup because my clusters are not named "cluster.local". Have you tried using `dnsConfig.searches`?

cognifloyd (changes requested):
This would break my setup, as, in my values:
- I include my private image registry in the name of all the images.
- I include `dnsConfig.searches` with the cluster namespace so that searches for the short name always resolve.

templates/_helpers.tpl
cognifloyd: Hmm. I guess my setup does not actually use this section, so it would not actually break my setup. I don't use the mongo/redis subcharts. But if I did install them in my cluster, this would prevent resolution because `cluster.local` does not exist.

So, I still wonder if adjusting `dnsConfig` would fix the issue for you.

templates/_helpers.tpl
```diff
@@ -165,7 +177,7 @@ Reduce duplication of the st2.*.conf volume details
 {{- if index .Values "mongodb" "enabled" }}
 {{- $mongodb_port := (int (index .Values "mongodb" "service" "port")) }}
 - name: wait-for-db
-  image: busybox:1.28
+  image: {{ template "stackstorm-ha.imageRegistry" . }}/library/busybox:1.28
```
cognifloyd, line 180: Hmm. I don't use the private registry for these - so something like this might make sense.

cognifloyd: What if we added a value: `image.utilityImage` that defaults to `busybox:1.28`. That would make upgrading that image easier, and also allow updating it to include the repository.

guzzijones: i can refactor to use that instead.

templates/_helpers.tpl
guzzijones: isn't `svc.cluster.local` a constant in k8?
yes i tried `dnsConfig.searches` and it didn't work.

guzzijones: i will add a values parameter for an optional cluster domain instead of 'cluster.local'

values.yaml
```diff
@@ -20,7 +22,7 @@ image:
   # See: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
   #pullSecret: "your-pull-secret"
 
-
+clusterDomain: cluster.local
```
cognifloyd, line 25: This needs some kind of doc string.

The clusterDomain is defined here: https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration
And the DNS server has to be configured with it as well:
https://stackoverflow.com/a/52941407/1134951

Looks like kube-dns is the thing that defaults to `cluster.local` (overridable with the `--domain` param): https://github.com/kubernetes/dns/blob/master/cmd/kube-dns/app/options/options.go#L59

I use coredns where there is not a default domain, and I did not use the `cluster.local` domain shown in the docs: https://coredns.io/plugins/kubernetes/

How is your cluster's dns configured?

values.yaml
```diff
@@ -5,6 +5,8 @@
 ## Docker image settings, applied to all StackStorm pods
 ##
 image:
+
+  utilityImage: ""
```
cognifloyd, line 9: This needs a doc string and a comment showing an example.

templates/_helpers.tpl
```diff
@@ -296,7 +308,7 @@ Merge packs and virtualenvs from st2 with those from st2packs images
   {{- if $.Values.st2.packs.images }}
     {{- range $.Values.st2.packs.images }}
 - name: 'st2-custom-pack-{{ printf "%s-%s-%s" .repository .name .tag | sha1sum }}'
-  image: "{{ .repository }}/{{ .name }}:{{ .tag }}"
+  image: "{{ .registry }}/{{ .repository }}/{{ .name }}:{{ .tag }}"
```
cognifloyd, line 311:
```suggestion
  image: "{{ .repository }}/{{ .name }}:{{ .tag }}"
```

I think you missed reverting this one.

values.yaml
guzzijones: I use coredns.
My configmap:
```.:53 {
    errors
    health
    kubernetes cluster.local in-addr.arpa ip6.arpa {
      pods insecure
      fallthrough in-addr.arpa ip6.arpa
    }
    prometheus :9153
    forward . /etc/resolv.conf
    cache 30
    loop
    reload
    loadbalance
}
```


```

Comment (guzzijones) [reply values.yaml:25]: on the node my kubelet-config.json does have a line for:
````
{
...
    clusterDomain": "cluster.local",
...
}
```

values.yaml
guzzijones: added a doc string

templates/_helpers.tpl
guzzijones: removed .registry parameter

values.yaml
guzzijones: I added a comment here as well.

cognifloyd (approved):

values.yaml
```diff
@@ -70,6 +81,17 @@ st2:
...
+    # fixes no replicaset found bug;  
+    [database]
+    # Connection retry backoff max (seconds).
+    connection_retry_backoff_max_s = 10
+    # Backoff multiplier (seconds).
+    connection_retry_backoff_mul = 1
+    # Connection retry total time (minutes).
+    connection_retry_max_delay_m = 3
+    # Connection and server selection timeout (in ms).
+    connection_timeout = 12000
```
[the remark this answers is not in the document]
guzzijones: I would agree here. the ideal solution would be to set the following in [dockerfiles](https://github.com/StackStorm/st2-dockerfiles/blob/dab5214ce7fe9897036f1dcb94bafa8eb42853b1/base/files/st2.tmp.conf#L17-L21)
```
    connection_timeout = 12000
```

Or some reasonable number and remove this change from this PR.
That way docker handles the restarts of the pods.

This PR could serve as documentation as to what the problem is with 'no replica sets found' and hopefully point people to the correct solution.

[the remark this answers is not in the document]
guzzijones: oh you want me to just have the connection_timeout because we override the other settings anyway. gotcha.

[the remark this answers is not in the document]
cognifloyd: The default timeout is 3s (3000 ms). If that needs to change, should it change in st2 core?

PR #357, opened 2023-02-02
Up the e2e-nightly timeout to 15m

mamercad:
Up the e2e-nightly timeout to 15m. Though, at the end of the day, I'm bothered that the variance is so high (and not even convinced that upping the timeout is the right thing to do). Consider this a placeholder for digging deeper.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
           command: helm dependency update
       - run:
           name: Helm install stackstorm-ha chart
-          command: helm install --timeout 10m0s --debug --wait --name-template stackstorm-ha .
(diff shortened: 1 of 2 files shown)


mamercad:
> image what's interesting it's `k8s-1.23.15` that fails.
> 
> https://kubernetes.io/releases/
> 
> > 1.23
> > Latest Release:1.23.16 (released: 2023-01-18)
> > End of Life:2023-02-28
> 
> One way is bumping the patch version `1.23.15` -> `1.23.16`, maybe it'll fix something. Another way is removing `1.23` from the build matrix as it's becoming EoL this month.

Good thoughts and observations, [heres's the PR](https://github.com/StackStorm/stackstorm-k8s/pull/358).

PR #359, opened 2023-02-07
Initial External AUTH Secret

bmarick:


bmarick:
I have done some initial testing already, but will do a more detailed testing tomorrow.

@cognifloyd  and @armab  -> I would appreciate your input on the name of the property to disable the Auth secret from being generated. I have no preference on the name of the property.

values.yaml
```diff
@@ -50,6 +50,12 @@ st2:
   # If set, st2.password always overrides any existing password.
   # If not set, the password is auto-generated on install and preserved across upgrades.
   #password: Ch@ngeMe
+  # Disable St2 auth secret generation
+  # If set, it requires a the customer to generate a secret with the name of "{{.Release.Name}}-st2-auth"
```
cognifloyd, line 54: Hmm. I was imagining the other direction.

Let the user specify the name of a secret they want to use instead of the `{{.Release.Name}}-st2-auth` secret. If they provide the name of an externally managed secret, then use that AND also skip generation of the `{{.Release.Name}}-st2-auth` secret.

bmarick:
FYI - I got pull off this work to do different tasks at my Job.
I am back and working to address the issues raised in this pr.
My hope is to have an update later today or tomorrow!

bmarick:
Please let me know what you think of the values file.
I would still like to add 1 test to validate the secret name in the deployment/secret get properly generated

values.yaml
bmarick: @cognifloyd 
I have pushed an update can you please review again?

Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 # StackStorm version which refers to Docker images tag
 appVersion: "3.8"
 name: stackstorm-ha
-version: 0.110.0
+version: 0.111.0-rc0
```
cognifloyd, line 5: Would you please remove this? We'll bump the version at release time.

values.yaml
```diff
@@ -61,6 +61,14 @@ st2:
   # If set, st2.password always overrides any existing password.
   # If not set, the password is auto-generated on install and preserved across upgrades.
   #password: Ch@ngeMe
+
+  ## Configure the st2 auth secret generated/used
+  auth_secret:
+    # Disable the creation of the secret to allow customers to manage the secret separately
+    create: true
+    # Set custom name of the secret for the ST2 Auth configuration
+    name: ""
```
cognifloyd, line 70: 2 settings for this is a little too complex I think. Is there a reason to have a bool separate from the name?

If the chart owns the secret, it should follow the chart's naming convention. If the user provides the name, then the chart should not touch that secret, just use it.

Maybe call the value `auth_secret_name`. If it is an empty string, then generate the secret name and generate the secret. If it is not empty, then just use the name as is and skip generating the secret.

templates/deployments.yaml
```diff
@@ -52,7 +52,7 @@ spec:
         {{- end }}
         envFrom:
         - secretRef:
-            name: {{ .Release.Name }}-st2-auth
+            name: {{- include "stackstorm-ha.secretSt2Auth" . }}
```
cognifloyd, line 55:
```suggestion
            name: {{ include "stackstorm-ha.secretSt2Auth" . }}
```

I don't think you need the `-` on these lines, do you?

templates/secrets_st2auth.yaml
```diff
@@ -4,12 +4,12 @@
 {{- if or $deprecated_username $deprecated_password }}
 {{- fail "Please update your values! Values for username and password moved from secrets.st2.* to st2.*" }}
 {{- end }}
+{{- if .Values.st2.auth_secret.create }}
```
cognifloyd, line 7: Based on https://stackoverflow.com/a/67718744 I think you can do:


```suggestion
{{- if not .Values.st2.auth_secret_name }}
```

Chart.yaml
bmarick: Done. That is used for my own testing, and why this pr is still in draft

values.yaml
bmarick: I adjusted to having a single value, like you suggested. I went with the name `authExistingSecret`. 
This is based on how Bitnami Helm charts appear to do it, [example](https://github.com/bitnami/charts/tree/main/bitnami/postgresql-ha#global-parameters).

Please let me know what you think

templates/deployments.yaml
bmarick: You are correct. After adding the additional testing, I found that that actually caused issues.

templates/secrets_st2auth.yaml
bmarick: That is correct, and similar to what I ended up doing

cognifloyd (commented):
LGTM. Let me know when you're ready for the final review. :)

bmarick:
> LGTM. Let me know when you're ready for the final review. :)

Will test today in my environment for real world testing. So request should come in later today

bmarick:
I have done live testing in my lab environment and confirmed this is functional

bmarick:
Okay I can confirm that the process fully works!

bmarick:
The only things missing are tests for the following files.
Please let me know if these are required.

- templates/jobs.yaml
- templates/tests/st2tests-pod.yaml

bmarick:
> LGTM. Let me know when you're ready for the final review. :)
> 

Ready for review!

tests/unit/secrets_test.yaml
```diff
@@ -0,0 +1,115 @@
...
+        enabled: true
+    release:
+      name: st2ha
+    asserts:
+      - hasDocuments:
+          count: 14
+      - equal:
+          path: spec.template.spec.initContainers[2].envFrom[0].secretRef.name
+          value: "hello-world"
+        documentIndex: 0
```
cognifloyd, line 86: Could you add a comment on the `documentIndex` for deployments and jobs to say which deployment/job this test is for?


Aside: I really wish we could use something other than index to select documents (and containers, ...) With helm-unittest. But oh well.

cognifloyd (commented):
Nice! One nitpick to save my sanity in the tests, and then I'll approve and merge.

Thank you for working on this!

tests/unit/secrets_test.yaml
bmarick: Resolved

bmarick:
The tests appear to have failed due to a rate limit

tests/unit/secrets_test.yaml
cognifloyd: oh. Sorry. I was not very specific. I was thinking of something more terse:
```
documentIndex: 0 # st2auth
```

If you look in the other test files, I've done something like that. Is that helpful or confusing for you? If adding this comment is confusing, then I can change how I do it in the other files for clarity.

bmarick: Done

cognifloyd (approved):

PR #363, opened 2023-03-08
allow datastore secret name to be specified

guzzijones:
allow a datastore secret name to be specified.  If one is specified do not dynamically generate a new datastore secret.
If I uninstall and reinstall this helm chart a new secret is generated.
Sometimes a helm upgrade isn't enough.  In that case I do an uninstall and reinstall.  This seems to be the case for ingress changes.

### DIFF
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -7,7 +7,12 @@
 {{- fail "Volume definition(s) missing! When st2.packs.volumes.enabled, you must define volumes for both packs and virtualenvs." }}
   {{- end }}
 {{- end }}
-
+{{- $crypto_key_name := print "no_name" }}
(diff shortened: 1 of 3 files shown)


guzzijones:
Any thoughts on this?

cognifloyd (changes requested):
I think we're going to need to do something like this for many of our secrets.

What about adding a helper that returns the name of the secret to use? That can handle returning the custom secret name if provided, or else the default name. Use that helper in the be deployments and jobs.

Then, in the secret file, wrap the whole resource in an if block. If there is a custom resource name, do not create the chart-managed secret.

cognifloyd:
Something like #359 would be great for the datastore secret.

guzzijones:
I am going to start on this with the changes you requested.

PR #366, opened 2023-05-15
add external secret for datastore encryption

guzzijones:
add an external secret check to values.yaml to allow for secret to be stored outside of helm chart.

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -35,7 +35,7 @@ jobs:
         # We should periodically check to see if another fork has taken over maintenance,
         # as the de-facto "best" fork has changed several times over the years.
         run: |
-          helm plugin install https://github.com/quintush/helm-unittest
+          helm plugin install https://github.com/quintush/helm-unittest --version v0.2.11
 
       - name: Install chart dependencies
         run: |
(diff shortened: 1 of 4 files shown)


guzzijones:
The helm-unittest has an invalid argument `--helm3`

cognifloyd:
Well ok then. Apparently, the `helm-unittest` fork we've been using is not the primary one any more :facepalm: . https://github.com/helm-unittest/helm-unittest/issues/102

OK. So, we need to switch to helm-unittest/helm-unittest but there are breaking changes that mean we have to change the cli arg (as you noticed) and some of our tests. One good thing is that the new version supports JSONPath, so we might be able to use names instead of indexes to verify things which will make our unit tests much less obtuse.

I'm not prepared to tackle all of that right now, so maybe we can pin a version on the old fork for the time being. Could you try changing https://github.com/StackStorm/stackstorm-k8s/blob/master/.github/workflows/unit.yaml#LL38C31-L38C72 like this:
```diff
-          helm plugin install https://github.com/quintush/helm-unittest
+          helm plugin install https://github.com/quintush/helm-unittest --version v0.2.11
```

cognifloyd (approved):

bmarick:
@cognifloyd - Any chance you can do a release so this can be available at https://helm.stackstorm.com/

PR #371, opened 2023-08-21
Remove redundant [credentials] header

### LINKED ISSUE #370: st2clientConfig issue with jobs.extra_hooks
We're trying to use extra_hooks in a k8s deployment and running into an issue where the additional jobs we define get run. But we find as we try to run additional st2 commands that we're seeing an error along the lines of: 

`configparser.DuplicateSectionError: While reading from '/root/.st2/config' [line 2]: section 'credentials' already exists`

In looking into this issue I see that at https://github.com/StackStorm/stackstorm-k8s/blob/415e3a672517762e747f446a13697464df210830/values.yaml#L925-L928 creating a st2clientConfig value that `[credentials]` is included in the config value. 
But in the jobs template '[credentials]' is being pre-prended to the st2clientConfig value: https://github.com/StackStorm/stackstorm-k8s/blob/415e3a672517762e747f446a13697464df210830/templates/jobs.yaml#L663-L666

resulting in a config file that looks like: 

```
[credentials]
[credentials]
username=st2admin
password=fred
```

cars:
Remove redundant [credentials] header for client config that causes st2 commands to fail.
Closes #370

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Reduce duplication in label tests (#354) (by @cognifloyd)
 * Add `st2canary` job as a Helm Hook that runs before install/upgrade to ensure `st2.packs.volumes` is configured correctly (if `st2.packs.volumes.enabled`). (#323) (by @cognifloyd)
 * Enable using existing `st2-auth` secret. This allows users to manage this secret outside of the Helm process. (#359) (by @bmarick)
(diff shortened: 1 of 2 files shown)


armab (changes requested):
Thanks for the contribution!

Could you please update the https://github.com/StackStorm/stackstorm-k8s/blob/master/CHANGELOG.md as well?

Otherwise looks good 👍

cars:
Changelog updated

armab (approved):

armab:
Thanks!

PR #374, opened 2023-09-07
add pod graceful timeout override

guzzijones:
add graceful timeout option for pods.
this will allow a user to override it in values.yaml.

we need this to to  along with the graceful timeout settings in st2actionrunner and st2workflowengine.

### DIFF
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -753,6 +753,7 @@ spec:
           {{- toYaml .Values.st2workflowengine.annotations | nindent 8 }}
         {{- end }}
     spec:
+      terminationGracePeriodSeconds: {{ .Values.st2workflowengine.terminationGracePeriodSeconds | default 300 }}
       imagePullSecrets:
       {{- if .Values.image.pullSecret }}
       - name: {{ .Values.image.pullSecret }}
@@ -1321,6 +1322,7 @@ spec:
       {{- if and .Values.st2.packs.images (not .Values.st2.packs.volumes.enabled) }}
         {{- include "stackstorm-ha.packs-initContainers" . | nindent 6 }}
       {{- end }}
+      terminationGracePeriodSeconds: {{ .Values.st2actionrunner.terminationGracePeriodSeconds | default 300 }}
       containers:
       - name: st2actionrunner
         {{- with .Values.st2actionrunner }}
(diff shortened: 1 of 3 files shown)


templates/deployments.yaml
```diff
@@ -1321,6 +1322,7 @@ spec:
       {{- if and .Values.st2.packs.images (not .Values.st2.packs.volumes.enabled) }}
         {{- include "stackstorm-ha.packs-initContainers" . | nindent 6 }}
       {{- end }}
+      terminationGracePeriodSeconds: {{ .Values.st2actionrunner.terminationGracePeriodSeconds | default 300 }}
```
armab, line 1325: The `300` default looks pretty much hidden from the user's perspective.

values.yaml
```diff
@@ -563,6 +563,7 @@ st2timersengine:
 # https://docs.stackstorm.com/reference/ha.html#st2workflowengine
 # Multiple st2workflowengine processes can run in active-active mode and will share the load and pick up more work if one or more of the processes become available.
 st2workflowengine:
+  #terminationGracePeriodSeconds: 350
```
armab, line 566:
```suggestion
  # include some small explanation about this setting
  terminationGracePeriodSeconds: 300
```
Could we set the default here instead and uncomment it? This way it will be easier to manage it, including controlling the defaults.

Please also add a small code comment so users know how it's helpful as it plays together with the st2.conf runner graceful shutdown, right?

armab (changes requested):
Very helpful!

Just requested a couple minor changes about defaults and docs for this setting.
Please also don't forget to include the Changelog.

guzzijones:
@armab I added the requested changes.

values.yaml
```diff
@@ -563,6 +563,8 @@ st2timersengine:
 # https://docs.stackstorm.com/reference/ha.html#st2workflowengine
 # Multiple st2workflowengine processes can run in active-active mode and will share the load and pick up more work if one or more of the processes become available.
 st2workflowengine:
+  # include some small explanation about this setting
```
armab, line 566: Needs some description in this placeholder describing the `terminationGracePeriodSeconds`, what it does and why it's needed.
Please include the same for `st2actionrunner.terminationGracePeriodSeconds`

guzzijones: added the explanation.

armab: Still seeing the old one.
Forgot to commit the changes?

guzzijones: added.

armab (approved):
Looks good, Thanks!

cognifloyd (approved):

PR #379, opened 2023-11-06
Update K8s to latest version

ZoeLeah:
The following changes have been made:

- Update to latest CircleCI orb versions
- Switch image from ubuntu-2204:2022.10.2 to ubuntu-2204:current to have always the latest ubuntu version
- Increase helm install timeout to 20 minutes
- Update to latest k8s and remove unsupported versions

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
           command: helm dependency update
       - run:
           name: Helm install stackstorm-ha chart
-          command: helm install --timeout 10m0s --debug --wait --name-template stackstorm-ha .
+          command: helm install --timeout 20m0s --debug --wait --name-template stackstorm-ha .
       - run:
           name: Helm test
           command: helm test stackstorm-ha
@@ -66,7 +66,6 @@ workflows:
                 - "v1.28.3"
                 - "v1.27.7"
                 - "v1.26.10"
-                - "v1.25.15"
   # Run periodic nightly Helm tests to ensure there are no regressions
   e2e-nightly:
     jobs:
@@ -78,7 +77,6 @@ workflows:
                 - "v1.28.3"
                 - "v1.27.7"
                 - "v1.26.10"
-                - "v1.25.15"
     triggers:
       - schedule:
           cron: "0 1 * * *"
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           command: helm dependency update
       - run:
           name: Helm install stackstorm-ha chart
-          command: helm install --timeout 10m0s --debug --wait --name-template stackstorm-ha .
+          command: helm install --timeout 20m0s --debug --wait --name-template stackstorm-ha .
```
armab, line 42: I didn't see any evidence that lifting install timeout helps.
Usually from the history of CI failures if deployment didn't complete in ~5mins, everything goes into a crash loop which is rarely recoverable without a manual intervention.
The usual reason behind the crash loop is one of the backend clusters didn't startup correctly like MongoDB, Redis or RabbitMQ.

But let's see if this helps.
Please use at least `15m` everywhere as bumping to `20mins` is too much!


```suggestion
          command: helm install --timeout 15m0s --debug --wait --name-template stackstorm-ha .
```

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## In Development
+* Bump to latest CircleCI orb versions (by @ZoeLeah)
+* Remove unsupported k8s Versions (by @ZoeLeah)
```
armab, line 5: Please mention explicitly (eg. 26,27,28) the new K8s versions that are used for testing this Helm chart.
That would provide a better info for someone reading the Changelog and thinking if particular K8s version is supported or not.

armab (changes requested):
Thanks for updates!

A couple of changes requested to address

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Changelog
 
 ## In Development
+* Bump to latest CircleCI orb versions (kubernetes@1.3.1 and helm@3.0.0 by @ZoeLeah)
+* Remove unsupported k8s Versions (1.24.x and 1.25.x by @ZoeLeah)
+* Update and add new K8s versions (1.28.3, 1.27.7 and 1.26.10 by @ZoeLeah)
+* Switch from ubuntu-2204:2022.10.2 to ubuntu-2204:current (by @ZoeLeah)
+* Update K3s to v1.28.3+k3s1 (by @ZoeLeah)
+* Increase helm install timeout to 15 minutes (by @ZoeLeah)
```
armab, line 9: That's too much of the changelog details for a single PR 😃 (balance is the key), but let's go with that

armab (approved):
Thanks for the K8s refresh!

PR #382, opened 2023-12-21
Presleep for st2auth, st2web, st2api, st2stream

guzzijones:
kubeproxy and kubelet race to shutdown and update firewall rules.  This PR gives kubeproxy a bit of time to update it's firewall rules.  
[pod deletes notify k8 api server and kubelet at the same time](https://blog.laputa.io/graceful-shutdown-in-kubernetes-85f1c8d586da)


mamercad (approved):
LGTM

PR #383, opened 2024-01-08
Update the Chart Maintainers - the StackStorm Authors

armab:
Update the chart Maintainers to be "the StackStorm Authors".
This is a common name used everywhere across the org since the LF transition: https://github.com/StackStorm/st2/blob/master/st2api/st2api/app.py#L1C17-L1C40

Shared contact would suit better, as seen in many other OSS projects/repos.


mamercad (approved):
🚢

PR #385, opened 2024-01-19
Secret DataStore Crypto Key should not be created when existing provided

bmarick:
Changes: 
- Update `secrets_datastore_crypto_key.yaml` to prevent creation when `st2.existingDatastoreSecret` is set
- Add unit tests to validate changes
- Fix unit test syntax

Addresses Issue #386


guzzijones (approved):
looks good to me.

PR #396, opened 2024-02-13
Added job to remove uninstalled packs from Mongo database

jk464:
Useful when changing the mounted pack image, that removes previously installed packs.

Normally the packs have been removed from disk, but are still registered in the Mongo database.

This runs a job as part of registering packs, to check for any packs no longer present on disk and removes them from the DB.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -390,6 +390,37 @@ Merge packs and virtualenvs from st2 with those from st2packs images
   {{- end }}
 {{- end -}}
 
+{{- define "stackstorm-ha.cleanup-packs" -}}
+  {{- if or $.Values.st2.packs.images $.Values.st2.packs.volumes.enabled }}
+# Clean Up packs - unregister packs no longer installed
+- name: st2-cleanup-packs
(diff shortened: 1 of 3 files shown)


cognifloyd (changes requested):
This would break my cluster. I use non-existent packs on purpose for ad-hoc rules and policies that a workflow I wrote creates when deploying an updated pack. So, they're very likely to exist when upgrading st2.

As a convention, I use dunder names (`__some_pack__`) for packs that are not expected to exist on disk. I use a policies to delay all of the actions in a pack during a pack upgrade. I'm planning to use ad-hoc rules for scheduling the run of some action at a later time.

I do agree that this is a problem. But, I suspect we need more primitives within st2 core to handle this in a cleaner way, and in a way that is not k8s-specific.

jk464:
@cognifloyd if you can give me any pointers into the `st2` core where we could add those new "primitives" I'd be happy to take a stab at re-implementing this functionality there (and I'd also add some logic to ignore dunder-name packs).

When I took a stab at this originally, I did look at doing it in core, but got lost trying to follow the code to locate the logic where pack loading happens at bootstrap - and therefore where I could add pack _un_ loading

PR #397, opened 2024-02-13
Allow templating of st2.packs.configs in values.yaml

jk464:
As it says on the tin.

Allows the use of go templating in the `st2.packs.configs` value when creating the Configs ConfigMap

### DIFF
--- a/templates/configmaps_packs.yaml
+++ b/templates/configmaps_packs.yaml
@@ -7,4 +7,4 @@ metadata:
     description: StackStorm pack configs defined in helm values, shipped in (or copied to) '/opt/stackstorm/configs/'
   labels: {{- include "stackstorm-ha.labels" (list $ "st2") | nindent 4 }}
 data:
-{{ toYaml .Values.st2.packs.configs | indent 2 }}
+{{ tpl (toYaml .Values.st2.packs.configs) . | indent 2 }}


cognifloyd:
Fascinating. Do you have any examples of how you would use this?

cognifloyd:
Would you look at adding some unit tests for `st2.packs.configs`? It is on this list of features that are missing unit tests: https://github.com/StackStorm/stackstorm-k8s/issues/28#issuecomment-1048530740

You can look at the other files under [tests/unit/](https://github.com/StackStorm/stackstorm-k8s/tree/master/tests/unit) to see how we've tested other features.

The tests will be a good way to show examples of how making the configs templatable could be used.

In particular I want to make sure something with a more complex template (maybe a range loop?) will render and get indented correctly. I wonder if we'll need to loop over each entry in `.Valus.st2.packs.configs` and template the contents of each file.

jk464:
> Would you look at adding some unit tests for `st2.packs.configs`? It is on this list of features that are missing unit tests: [#28 (comment)](https://github.com/StackStorm/stackstorm-k8s/issues/28#issuecomment-1048530740)
> 
> You can look at the other files under [tests/unit/](https://github.com/StackStorm/stackstorm-k8s/tree/master/tests/unit) to see how we've tested other features.
> 
> The tests will be a good way to show examples of how making the configs templatable could be used.
> 
> In particular I want to make sure something with a more complex template (maybe a range loop?) will render and get indented correctly. I wonder if we'll need to loop over each entry in `.Valus.st2.packs.configs` and template the contents of each file.

I can take a stab at it - but could we get https://github.com/StackStorm/stackstorm-k8s/pull/417 reviewed and merge first so I don't have to write the tests twice ;)

PR #399, opened 2024-02-13
Move st2tests to a job

jk464:
Instead of running `st2tests` as a `pod`, this runs it as a `job`.

To ensure the job only runs once, and fails if a test fails, I've added:

```
      restartPolicy: Never
  backoffLimit: 0
```

Which means it'll never Back off / or retry.


templates/_helpers.tpl
```diff
@@ -191,6 +191,23 @@ Reduce duplication of the st2.*.conf volume details
...
 
+{{- define "stackstorm-ha.init-containers-wait-for-auth" -}}
+- name: wait-for-auth
+  image: "alpine/curl:8.4.0"
+  imagePullPolicy: {{ .Values.image.pullPolicy }}
+  command:
+    - 'sh'
+    - '-c'
+    - >
+      until curl -skSL --fail -w '\n'  -X POST -u {{ .Values.st2.username }}:{{ .Values.st2.password }} "https://{{ required ".Values.ingress.fqdn is required if .Values.ingress.class is non-empty" .Values.ingress.fqdn | printf (ternary "canary-%s" "%s" .Values.phaseCanary)}}/auth/tokens"; do
```
cognifloyd, line 202: I don't see these vars in our values file:
- `ingress.fqdn`
- `ingress.class`
- `phaseCanary`

https://github.com/StackStorm/stackstorm-k8s/blob/17e5fca27a38164d77160e341ce8ba8d531f50e0/values.yaml#L298-L321

I also do not use the ingress, so these would never be defined for me.

user/pass is also only available in my cluster(s) during initial bootstrap. Once I switch to ldap, I exclusively use api tokens within helm.

Do you have any ideas on how to make this more generic?

cognifloyd (changes requested):
It's hard to see what you've added in the github interface, so I'm highlighting the new sections in my review to help myself see them.

It looks like you want to use the helm tests as a canary script that you run after install or upgrade. Is that right? Are you already using something like this? How well does it work for you?

jk464:
I think we can just drop this PR, I've made a misunderstanding on how these `tests` work, and that they for use with the `helm test` command.

Internally we use `helm template` and then `kubectl apply` to deploy helm charts to our `k8s` cluster - this seemingly had the side effect of deploying the `tests` `pod` as part of the deploy - which AFAIK is now a mistake on _our_ part and us misusing this `pod`.

Please let me know if I've yet again misunderstood how this is _meant_ to work...

cognifloyd:
> I think we can just drop this PR, I've made a misunderstanding on how these `tests` work, and that they for use with the `helm test` command.

Yeah. `helm test` is the primary use case for this. Having some kind of canary job might be interesting if you have an idea of what that job should do.

So far, I've added one `st2canary` job to help catch common issues around packs volumes:

https://github.com/StackStorm/stackstorm-k8s/blob/93e3f265b451acd00e1bd71c09dde2b7b3c4397d/templates/jobs.yaml#L491-L604

Any ideas on potential `st2canary` jobs?

> Internally we use `helm template` and then `kubectl apply` to deploy helm charts to our `k8s` cluster - this seemingly had the side effect of deploying the `tests` `pod` as part of the deploy - which AFAIK is now a mistake on _our_ part and us misusing this `pod`.

Ah yes. That's a common gotcha (at least where I work). I've been encouraging my coworkers to use `helm install` and `helm upgrade` because doing `helm template` bypasses a lot of the safety checks helm does during install/upgrade. To me, `helm` seems a bit more graceful than `kubectl apply`.

That said, I wonder if there is a way to exclude the helm test hooks when running `helm template`? Or maybe we can skip rendering the tests stuff in `helm template` with a new value (or maybe helm exposes some metadata we can use here)? Please submit a PR if there's something the chart can do to avoid the `helm template` gotcha.

> Please let me know if I've yet again misunderstood how this is _meant_ to work...

:smile: You've got it now. The idea of canary jobs is very intriguing though...

Issue #245, opened 2021-09-09
Make packs/virtualenv file copy more robust


cognifloyd:
For system packs, and for any images in `st2.packs.images`, we copy packs and virtualenvs into /opt/stackstorm.

Before #199, the packs volumes were always `emptyDir`, so using `/bin/cp -aR` was just fine.
In https://github.com/StackStorm/stackstorm-ha/pull/160#issuecomment-734927852, it was noted that the `st2packs` image does not have `rsync` so we have continued with `/bin/cp` even though rsync could be a bit better.

After #199, the underlying volume's filesystem might not support preserving ownership, timestamps or other attributes, as reported in #238. In #238, the NFS exports were setup with `root_squash` which prevented ownership preservation. Debugging that was quite annoying.

So, it would be better to handle different filesystems more gracefully. If we can preserve ownership and file permissions, great. But we should only fail if the file contents do not get copied. We might also need to fail if executable files are not executable after copy.

`rsync` is probably the best tool to do this -> We need to add it to the `st2packs` image.
But, we do not want to force everyone to rebuild their `st2packs` images, especially if they are not and do not intend to use the new `st2.packs.volumes` feature. So, if `rsync` is not available, we need to fall back to using `cp`.

See comments:
- [a description of where we need to adjust the chart](https://github.com/StackStorm/stackstorm-ha/issues/238#issuecomment-914663644)
- [nfs export options](https://github.com/StackStorm/stackstorm-ha/issues/238#issuecomment-914672590)
- [adding rsync to `st2packs`](https://github.com/StackStorm/stackstorm-ha/issues/238#issuecomment-915244500)
- [reply to adding rsync to `st2packs`](https://github.com/StackStorm/stackstorm-ha/issues/238#issuecomment-915258055)

We should probably do this before prod/GA to have minimal impact on GA adopters.

PR #400, opened 2024-02-13
Run Containers as Non-Root, and without Privilege Escalation by default.

jk464:
In attempt to harden the security of running StackStorm within k8s, this PR makes it such that by default all containers run as a non-root user, generally `1000:1000` (the `stanley` user) - or for `st2web` it runs as `100:100` (which is the `nginx` user).

It also disables the ability of the container to escalate privileges.

 Generally in my testing, this all "just" works - Aside from requiring a change to the `st2web` container to allow `nginx` to run as the `nginx` user. (Changes to support that are here https://github.com/StackStorm/st2-dockerfiles/pull/66).

NOTE: This PR depends on https://github.com/StackStorm/st2-dockerfiles/pull/66 to work.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -342,8 +342,8 @@ Merge packs and virtualenvs from st2 with those from st2packs images
     - 'sh'
     - '-ec'
     - |
-      /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
-      /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+      /bin/cp -dR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
+      /bin/cp -dR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
   {{- with $.Values.securityContext }}
   securityContext: {{- toYaml . | nindent 8 }}
   {{- end }}
@@ -363,8 +363,8 @@ Merge packs and virtualenvs from st2 with those from st2packs images
     - 'sh'
     - '-ec'
     - |
-      /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
-      /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+      /bin/cp -dR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
+      /bin/cp -dR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
   {{- with .Values.securityContext }}
   securityContext: {{- toYaml . | nindent 8 }}
   {{- end }}
@@ -383,7 +383,7 @@ Merge packs and virtualenvs from st2 with those from st2packs images
     - 'sh'
     - '-ec'
     - |
-      /bin/cp -aR /opt/stackstorm/configs/. /opt/stackstorm/configs-shared
+      /bin/cp -dR /opt/stackstorm/configs/. /opt/stackstorm/configs-shared
   {{- with .Values.securityContext }}
   securityContext: {{- toYaml . | nindent 8 }}
   {{- end }}
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -432,13 +432,13 @@ spec:
         securityContext: {{- toYaml . | nindent 10 }}
         {{- end }}
         ports:
-        - containerPort: {{ eq (get .Values.st2web.env "ST2WEB_HTTPS" | toString) "1" | ternary 443 80 }}
+        - containerPort: {{ eq (get .Values.st2web.env "ST2WEB_HTTPS" | toString) "1" | ternary 8443 8080 }}
         # Probe to check if app is running. Failure will lead to a pod restart.
         livenessProbe:
           httpGet:
             scheme: {{ eq (get .Values.st2web.env "ST2WEB_HTTPS" | toString) "1" | ternary "HTTPS" "HTTP" }}
             path: /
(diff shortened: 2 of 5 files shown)


templates/_helpers.tpl
```diff
@@ -342,8 +342,8 @@ Merge packs and virtualenvs from st2 with those from st2packs images
     - 'sh'
     - '-ec'
     - |
-      /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
-      /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+      /bin/cp -dR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
+      /bin/cp -dR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
```
cognifloyd, line 346: `-d` only preserves symlinks as-is. `-a` also copied the mode, including execute bits.

We need to make sure to preserve the execute bit on files. Would we have to add `--preserve=mode` to copy that?

Also, how will this interact with #245 (Switch from `cp` to `rsync` with fallback to `cp`)?

templates/deployments.yaml
```diff
@@ -1614,13 +1614,13 @@ spec:
         {{- end }}
         volumeMounts:
         - name: st2client-config-vol
-          mountPath: /root/.st2/
+          mountPath: /home/stanley/.st2/
```
cognifloyd, line 1617: Is the switch from `/root` to `/home/stanley` covered in the docker image changes?

We need to support older docker images too. So maybe we need to pass this path in via values?

templates/services.yaml
```diff
@@ -100,7 +100,7 @@ spec:
   {{- end }}
   ports:
   - protocol: TCP
-    port: {{ eq (get .Values.st2web.env "ST2WEB_HTTPS" | toString) "1" | ternary 443 80 }}
+    port: {{ eq (get .Values.st2web.env "ST2WEB_HTTPS" | toString) "1" | ternary 8443 8080 }}
```
cognifloyd, line 103: We might need this to be in values to support older docker images.

templates/deployments.yaml
cognifloyd: Apparently st2.conf is not embedded in the Docker image like I thought: https://github.com/StackStorm/st2-dockerfiles/blob/master/base/files/st2.docker.conf

And `st2.conf` is irrelevant here any way, as this is the path to the `~/.st2` that you are adjusting, not `~/.ssh`. oops.

I wonder if this needs to be based on values:
```suggestion
          {{- if .Values.st2.system_user.user == 'root' }}
          mountPath: /root/.st2/
          {{- else }}
          mountPath: /home/{{ .Values.st2.system_user.user }}/.st2/
          {{- end }}
```

The assumptions here are spelled out in Values:
https://github.com/StackStorm/stackstorm-k8s/blob/17e5fca27a38164d77160e341ce8ba8d531f50e0/values.yaml#L118-L121

We could also add a new value `st2.system_user.home`. wdyt?

values.yaml
```diff
@@ -290,8 +290,11 @@ st2:
 ## Default SecurityContext for pods and containers.
 ## Overrides available for st2web, st2actionrunner, st2sensorcontainer, and st2client pods.
 ##
-podSecurityContext: {}
-securityContext: {}
+podSecurityContext:
+  runAsNonRoot: true
+securityContext:
+  runAsUser: 1000
+  allowPrivilegeEscalation: false
```
cognifloyd, line 297: In my cluster, I use drop capabilities, re-adding them only for the deployments that need it. So I have:

```yaml
#podSecurityContext:
securityContext:
  capabilities:
    drop: ["ALL"]
st2actionrunner:
  #podSecurityContext:
  securityContext:
    capabilities:
      drop: ["ALL"]
      add:
        - chown
        - fowner # chmod
        - fsetid
        - setfcap
        - setgid
        - setuid
        - dac_override # allows root to bypass "discretionary access control" (file rwx permission checks)
        - kill
        - net_raw # ping
        - audit_write # sudo
        - setpcap
st2web:
  #podSecurityContext:
  securityContext:
    capabilities:
      drop: ["ALL"]
      add:
        - chown
        - setgid
        - setuid
        - dac_override # allows root to bypass "discretionary access control" (file rwx permission checks)
        - net_bind_service # bind privileged ports 80 or 443
```

So, that's a slightly different approach than just running as a non-root user. In particular, if st2web is not running as root, and is using unprivileged ports, then the config could be somewhat simpler.

templates/_helpers.tpl
cognifloyd: Would you please check out #415? I'd like to come up with the minimal set of flags for `cp` and `rsync`.

jk464: @Stealthii made these changes (which remains me I should credit him as an author in the commits..), I believe this was to fix issues with running as a non-privileged user. @Stealthii can you comment here / have a look at the above PRs :D

templates/deployments.yaml
jk464: I like the check you have for `root`, so I've just gone ahead and taken that

values.yaml
jk464: I think for `st2web`, the fact I don't bind privileged ports, and also I ensured any required files are `chown` by `nginx` in the Docker image removes any need for this capabilities?

For `st2actionrunner` do we have any documentation that points towards what `st2actionrunner` needs each of those capabilities for? I'm wondering if we can remove any of them by making changes elsewhere, w/o breaking compatibility ofc.

templates/_helpers.tpl
Stealthii: @cognifloyd I've commented on #415 with a proposed change that should preserve all relevant attributes besides ownership. https://github.com/StackStorm/stackstorm-k8s/pull/415#issuecomment-2100894943

Once verified and if it becomes part of that PR, we can drop the change here and have this PR focus solely on the enablement of non-root runtimes.

PR #401, opened 2024-02-13
Enabled internal TLS between k8s pods by default

jk464:
Continuing on from the theme in (https://github.com/StackStorm/stackstorm-k8s/pull/400) this PR is another attempt at improving the security hardening of StackStorm in `k8s`.

By default - all inter-pod communication between pods within a `k8s` deployment use non-encrypted protocols. (i.e. Between `st2` and `Mongo` / `RabbitMQ` and `Redis`, and to the internal `api`, `auth` and `stream` endpoints within `st2`)

This is fine if your security model is hard (secure) on the outside (public facing), but crunchy (insecure) on the inside (private facing) - (like an [Armadillo](https://www.youtube.com/watch?v=bqeGxMgVOHI) 😉).

However, ideally a modern approach would be zero-trust and secure communication everywhere.

This PR therefore, enables by default, Encryption/TLS everywhere*. It makes use of [cert-manager](https://cert-manager.io/) to enabled automatic generation of certificates, either using a user provided CA (to allow use of a corporate CA) or if the user already has a cert issuer setup - one can provide that.

TLS is enabled/configured globally for `st2` endpoints (`api`, `auth`, `stream`) under the `st2` block in `values.yaml`

```
  tls:
    enabled: true
    secretName: "internal-tls"
    mountPath: "/etc/ssl/internal"
    certificate_issuer:
      existing: false
      name: stackstorm-issuer
```

And then the individual endpoints can be toggled on/off under their respective blocks in `values.yaml`.

Sadly the `mongo` and `rabbitmq` TLS can't be setup from the `st2` config directly - so they're present (and have defaults configured) under their respective blocks as well.

The `st2` templates have then been updated to automatically configure all the required settings to use these certificates to enabled TLS - and the cert/key/ca are also automatically mounted to all the containers. 

Note since the `api` and `stream` endpoints don't support encryption, we have added a TLS proxy layer (using [ghostunnel](https://github.com/ghostunnel/ghostunnel)) infront of them to enable _inter_-pod encryption while allowing unencrypted communication on the _intra_-pod commuication between the `st2` component and the proxy layer.

*Note that the [OpenStack Tooz](https://github.com/openstack/tooz) library, which `st2` uses to talk to Redis, doesn't have support to configure the settings needed to enable TLS when using Sentinel - so Redis traffic is still unencrypted in this PR. However, we have an internal build of `Tooz` that adds this support, and we've also got the changes to the StackStorm helm template to support that working internally. Once we have our changes to `Tooz` merged upstream, we can provide a PR to `stackstorm-k8s` and `st2` core library to enable this support as well.

### DIFF
--- a/templates/ca.yaml
+++ b/templates/ca.yaml
@@ -1,0 +1,21 @@
+{{- if not ( .Values.st2.tls.certificate_issuer.existing ) -}}
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: {{ .Values.st2.tls.certificate_issuer.name }}
+  namespace: "{{ $.Release.Namespace }}"
+spec:
+  ca:
+    secretName: {{ .Values.st2.tls.certificate_issuer.name }}-tls
+---
+apiVersion: v1
+data:
+  tls.crt: "{{ .Values.secret.ca.crt }}"
+  tls.key: "{{ .Values.secret.ca.key }}"
+kind: Secret
+metadata:
+  name: {{ .Values.st2.tls.certificate_issuer.name }}-tls
+  namespace: "{{ $.Release.Namespace }}"
+type: kubernetes.io/tls
+{{- end -}}
--- a/templates/configmaps_st2-urls.yaml
+++ b/templates/configmaps_st2-urls.yaml
@@ -7,6 +7,18 @@ metadata:
     description: StackStorm service URLs, used across entire st2 cluster
   labels: {{- include "stackstorm-ha.labels" (list $ "st2") | nindent 4 }}
 data:
+  {{- if and .Values.st2.tls.enabled .Values.st2auth.tls.enabled }}
+  ST2_AUTH_URL: https://{{ .Release.Name }}-st2auth:9100/
+  {{- else }}
   ST2_AUTH_URL: http://{{ .Release.Name }}-st2auth:9100/
+  {{- end }}
+  {{- if and .Values.st2.tls.enabled .Values.st2api.tls.enabled }}
+  ST2_API_URL: https://{{ .Release.Name }}-st2api:9111/
+  {{- else }}
   ST2_API_URL: http://{{ .Release.Name }}-st2api:9101/
+  {{- end }}
+  {{- if and .Values.st2.tls.enabled .Values.st2stream.tls.enabled }}
+  ST2_STREAM_URL: https://{{ .Release.Name }}-st2stream:9112/
+  {{- else }}
   ST2_STREAM_URL: http://{{ .Release.Name }}-st2stream:9102/
+  {{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -88,10 +88,41 @@ st2:
     [api]
     allow_origin = '*'
     # fixes no replicaset found bug;
+
     [database]
     # Connection and server selection timeout (in ms).
     connection_timeout = 5000
+    {{- if .Values.mongodb.tls.enabled }}
+    ssl = True
+    ssl_ca_certs = {{ .Values.st2.tls.mountPath }}/ca.crt
+    ssl_cert_reqs = optional
+    ssl_certfile = {{ .Values.st2.tls.mountPath }}/tls.crt
+    ssl_keyfile = {{ .Values.st2.tls.mountPath }}/tls.key
+    {{- else }}
+    ssl = False
+    {{- end }}
 
+    [messaging]
+    {{- if .Values.rabbitmq.tls.enabled }}
+    ssl = True
+    ssl_ca_certs = {{ .Values.st2.tls.mountPath }}/ca.crt
+    ssl_cert_reqs = optional
+    ssl_certfile = {{ .Values.st2.tls.mountPath }}/tls.crt
+    ssl_keyfile = {{ .Values.st2.tls.mountPath }}/tls.key
+    {{- else }}
+    ssl = False
+    {{- end }}
+
+    [auth]
+    {{- if and .Values.st2.tls.enabled .Values.st2auth.tls.enabled }}
+    use_ssl = True
+    key = {{ .Values.st2.tls.mountPath }}/tls.key
+    cert = {{ .Values.st2.tls.mountPath }}/tls.crt
+    debug = False
+    enable = True
+    {{- else }}
+    use_ssl = False
+    {{- end }}
 
   #Override Definitions can be added here.
   #https://docs.stackstorm.com/latest/packs.html#overriding-pack-defaults
@@ -286,6 +317,15 @@ st2:
       #  roles:
       #    - "admin"
 
+  # Controls configuring TLS between internal inter-pod communications
+  tls:
+    enabled: true
+    secretName: "internal-tls"
+    mountPath: "/etc/ssl/internal"
+    certificate_issuer:
+      existing: false
+      name: stackstorm-issuer
+
 ##
 ## Default SecurityContext for pods and containers.
 ## Overrides available for st2web, st2actionrunner, st2sensorcontainer, and st2client pods.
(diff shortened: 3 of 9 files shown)


templates/configmaps_st2-urls.yaml
```diff
@@ -7,6 +7,18 @@ metadata:
     description: StackStorm service URLs, used across entire st2 cluster
   labels: {{- include "stackstorm-ha.labels" (list $ "st2") | nindent 4 }}
 data:
+  {{- if and .Values.st2.tls.enabled .Values.st2auth.tls.enabled }}
+  ST2_AUTH_URL: https://{{ .Release.Name }}-st2auth:9100/
+  {{- else }}
   ST2_AUTH_URL: http://{{ .Release.Name }}-st2auth:9100/
```
cognifloyd, line 13: We can reduce some of the repetition here by front loading the calculation:
```suggestion
  {{- if and .Values.st2.tls.enabled .Values.st2auth.tls.enabled }}
    {{- $proto := "https" }}
  {{- else }}
    {{- $proto := "http" }}
  {{- end }}
  ST2_AUTH_URL: {{ $proto }}://{{ .Release.Name }}-st2auth:9100/
```
and the same for the others.

values.yaml
```diff
@@ -88,10 +88,41 @@ st2:
...
+    [auth]
+    {{- if and .Values.st2.tls.enabled .Values.st2auth.tls.enabled }}
+    use_ssl = True
+    key = {{ .Values.st2.tls.mountPath }}/tls.key
+    cert = {{ .Values.st2.tls.mountPath }}/tls.crt
+    debug = False
+    enable = True
+    {{- else }}
+    use_ssl = False
+    {{- end }}
```
cognifloyd, line 125: This generated stuff should go in `st2.docker.conf` (in `templates/configmaps_st2-conf.yaml`)

values.yaml
```diff
@@ -370,7 +410,19 @@ st2web:
...
   ## Note that Helm templating is supported in 'mount' and 'volume'
-  extra_volumes: []
+  extra_volumes:
+    - name: st2-cert
+      mount:
+        mountPath: "/etc/ssl/st2/"
+        readOnly: true
+      volume:
+        secret:
+          secretName: "stackstorm-tls{{if .Values.phaseCanary}}-canary{{end}}"
```
cognifloyd, line 420: phaseCanary is not defined in our values file.

templates/configmaps_st2-urls.yaml
cognifloyd: oh. I see you're changing the ports for st2api and st2stream as well. Hmm.

st2auth already exposes the tls bits. Could we add tls to st2api and st2stream as well to avoid adding another process to the mix?

templates/ca.yaml
cognifloyd, line 1: Fascinating. I should look into installing cert-manager again to see if that would be work in my cluster.

Right now, I'm using this chart to generate a CA+cert that I load into the st2web pods:
https://github.com/cognifloyd/helm-gen-self-signed-cert/

`helm repo add self-signed-cert https://cognifloyd.github.io/helm-gen-self-signed-cert/`

templates/configmaps_st2-urls.yaml
jk464: Sorry I don't follow - do you mean enhance the `st2apit` and `st2stream` libraries themselves to support TLS out-of-the-box instead of having to stick `stunnel` infront of them on the `k8s` end?

values.yaml
jk464: Whoops, that was an internal change I missed removing - fixed it now

templates/ca.yaml
jk464: It works well for us, only issue we encounter is using _self-signed_ certs, doesn't work - `cert-manager` seems to forcibly sets `isCA` false in the certificate (regardless if you tell it not to) which breaks providing the self-signed cert as a CA, to self validate the cert.

Instead you have to do what we do here and create a internal CA and perform "proper" cert signing and provide the CA then. Although this does have the benefit for corporate users where if they have a corporate CA, they can use that and have "correctly" signed certs - so  I guess its a better implementation in the end.

I did want to avoid adding a hard requirement for `cert-manager` - so I've left that as an optional feature, although I would recommend using it, since `cert-manager` is open source and has no associated cost and does massively simplify certificate management 😄

jk464: Scratch that I forgot what I did 😅  - this _does_ require `cert-manager` nvm

templates/configmaps_st2-urls.yaml
cognifloyd: > Sorry I don't follow - do you mean enhance the `st2apit` and `st2stream` libraries themselves to support TLS out-of-the-box instead of having to stick `stunnel` infront of them on the `k8s` end?

Yes. `st2api` supports tls, so `st2api` and `st2stream` should do the same.

cognifloyd:
```suggestion
  {{- $proto := "http" }}
  {{- $port := "9100" }}
  {{- if and .Values.st2.tls.enabled .Values.st2auth.tls.enabled }}
    {{- $proto := "https" }}
    {{- $port := "9110" }}
  {{- end }}
  ST2_AUTH_URL: {{ $proto }}://{{ .Release.Name }}-st2auth:{{ $port }}/
```

values.yaml
```diff
@@ -1050,6 +1083,13 @@ mongodb:
   arbiter:
     enabled: false
   resources: {}
+  tls:
+    enabled: true
+    replicaset:
+      existingSecrets:
+        - internal-tls
+        - internal-tls
+        - internal-tls
```
cognifloyd, line 1092: Why the duplication here?

values.yaml
```diff
@@ -380,7 +389,19 @@ st2web:
   securityContext: {}  # NB: nginx requires some capabilities, drop ALL will cause issues.
   # mount extra volumes on the st2web pod(s) (primarily useful for k8s-provisioned secrets)
   ## Note that Helm templating is supported in 'mount' and 'volume'
-  extra_volumes: []
+  extra_volumes:
```
cognifloyd, line 392: extra_volumes is only for user-provided volumes so that if someone sets it to an empty list it doesn't break chart functionality.

values.yaml
jk464: MongoDB helm template requires an `existingSecret` for _each_ mongodb node, I assume with the intention that each node has its own cert - I've however instead created a cert thats valid for all nodes and just use the same one for all nodes - however, the Mongo chart still requires all 3 to be explicitly defined.

templates/configmaps_st2-urls.yaml
cognifloyd: So, it looks like we need to copy the st2auth ssl config opts to st2api and st2stream.

st2auth seems to define the ssl bits here:
- [st2auth/st2auth/config.py](https://github.com/StackStorm/st2/blob/180deff2f27f23da65e253a46f933365e40faf0f/st2auth/st2auth/config.py)
- [st2auth/st2auth/app.py](https://github.com/StackStorm/st2/blob/180deff2f27f23da65e253a46f933365e40faf0f/st2auth/st2auth/app.py)
- [st2auth/st2auth/cmd/api.py](https://github.com/StackStorm/st2/blob/180deff2f27f23da65e253a46f933365e40faf0f/st2auth/st2auth/cmd/api.py)

The SSL bits would need to go in these st2api files:
- [st2api/st2api/config.py](https://github.com/StackStorm/st2/blob/180deff2f27f23da65e253a46f933365e40faf0f/st2api/st2api/config.py)
- [st2api/st2api/app.py](https://github.com/StackStorm/st2/blob/180deff2f27f23da65e253a46f933365e40faf0f/st2api/st2api/app.py)
- [st2api/st2api/cmd/api.py](https://github.com/StackStorm/st2/blob/180deff2f27f23da65e253a46f933365e40faf0f/st2api/st2api/cmd/api.py)

The SSL bits would need to go in these st2stream files:
- [st2stream/st2stream/config.py](https://github.com/StackStorm/st2/blob/180deff2f27f23da65e253a46f933365e40faf0f/st2stream/st2stream/config.py)
- [st2stream/st2stream/app.py](https://github.com/StackStorm/st2/blob/180deff2f27f23da65e253a46f933365e40faf0f/st2stream/st2stream/app.py)
- [st2stream/st2stream/cmd/api.py](https://github.com/StackStorm/st2/blob/180deff2f27f23da65e253a46f933365e40faf0f/st2stream/st2stream/cmd/api.py)

jk464:
@cognifloyd thanks for the additional comments, I'll take a look at them on Monday, and I also have a go at updating st2api and stream to add TLS support then aswell 😊

templates/configmaps_st2-urls.yaml
jk464: Hey @cognifloyd I've opened https://github.com/StackStorm/st2/pull/6204 to add SSL/TLS support to `api` and `stream` if you could take a look that would be great :)

PR #403, opened 2024-02-16
Add missing policy and fix function

skiedude:
Fixes for https://github.com/StackStorm/stackstorm-k8s/issues/388

### DIFF
--- a/templates/jobs.yaml
+++ b/templates/jobs.yaml
@@ -540,7 +540,7 @@ spec:
             and then delete this st2canary pod manually.
             INTRO
 
-            function __handle_error__(){
+            function __handle_error__ {
               cat <<- 'FAIL'
               ERROR: One or more volumes in st2.packs.volumes (from helm values) does not meet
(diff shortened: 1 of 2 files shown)


templates/jobs.yaml
```diff
@@ -577,6 +577,7 @@ spec:
           {{- toYaml $.Values.st2.packs.volumes.configs | nindent 10 }}
         {{- end }}
         # st2canary job does not support extra_volumes. Let us know if you need this.
+      restartPolicy: OnFailure
```
copart-jafloyd, line 580:
```suggestion
      restartPolicy: Never
```

If the test pod fails, it should not restart -- it should just stay in the failed state.

cognifloyd, line 580:
```suggestion
      restartPolicy: Never
```

If the test pod fails, it should not restart -- it should just stay in the failed state.

cognifloyd (approved):

Issue #388, opened 2024-01-25
job-ensure-packs-volumes-are-writable missing policy and syntax error


skiedude:
**Using the current master branch**

When configuring the use of ReadWriteMany volumes the job in `templates/job.yaml` 
```name: {{ $.Release.Name }}-job-ensure-packs-volumes-are-writable```
Is fired off before other pods are started. This job fails to start because of 
```
  Error: failed pre-install: warning: Hook pre-install stackstorm-ha/templates/jobs.yaml failed: 1 error occurred:
  	* Job.batch "stackstorm-job-ensure-packs-volumes-are-writable" is invalid: spec.template.spec.restartPolicy: Required value: valid values: "OnFailure", "Never"
 ```
 
 It is missing the `restartPolicy: OnFailure` definition. 

Once that is in place, the job then fails for a sytax error 
```
╰─> kubectl logs stackstorm-job-ensure-packs-volumes-are-writable-lvc4k
Testing write permissions for packs volumes.
If this passes, the pod will automatically be deleted.
If this fails, inspect the pod for errors in kubernetes,
and then delete this st2canary pod manually.
sh: 8: Syntax error: "(" unexpected
```

This is in reference to `function __handle_error__()` in the `command` definition. This was temporarily fixed by just removing all the 
```
cat << 
cat <<-
```
calls and just keep the for loop. 

I can submit a PR for the missing `restartPolicy`, however the cat statements are something I'm unfamiliar with.

skiedude:
> @skiedude how are you getting past this? I am implementing custom st2 packs mounted via NFS storage account and facing the same issue. Im using the v1.0.0 stable charts

First thing you need to do is add `restartPolicy: OnFailure` to https://github.com/StackStorm/stackstorm-k8s/blob/master/templates/jobs.yaml#L564 with the same indentation level of the `volumes`

Then you will need to comment out lines https://github.com/StackStorm/stackstorm-k8s/blob/master/templates/jobs.yaml#L536-L555 as the cat/function stuff is broken.

cognifloyd:
To fix the syntax issue, this is probably the required change:

```diff
-            function __handle_error__(){
+            function __handle_error__ {
```

PR welcome

yoadev:
This bug is still present on master
Can we reopen this issue, i have a PR to fix all this script

cognifloyd:
@yoadev What errors are you seeing?

PR #407, opened 2024-03-19
solved issue #388 adding the restartPolicy and removing the write permissions check for the packs volumes

cayohollanda:


rafa13bb:
Resolved the issue locally in my tests

cognifloyd (changes requested):
I'm not ok with removing the error message and instructions on how to resolve issues with this.

Do you still see these errors now that #403 has been merged? What error messages are you seeing?

templates/jobs.yaml
```diff
@@ -573,6 +552,7 @@ spec:
           readOnly: false
         # TODO: Find out default resource limits for this specific job (#5)
         #resources:
+      restartPolicy: OnFailure
```
cognifloyd, line 555:
```suggestion
```

This was added in #403. You can see it below on line 584 below.

https://github.com/StackStorm/stackstorm-k8s/blob/fabbea9fb89bce6d093ad718def3e28fcfc47c5e/templates/jobs.yaml#L582-L586

fdrab: one more thing required for this to not fail is to remove the 2 leading whitespaces before FAIL on line 557. unless someone does it before me, I'll submit a PR once my company approves me signing of the CLA

cognifloyd: Cool. I have a fix for that. I'll push a PR in a few minutes.

cognifloyd: Fix pushed in #411

PR #410, opened 2024-04-10
add securityContext for more containers with better fallbacks

cognifloyd:
This adds values to define the `securityContext` for:
- custom st2packs images
- extra_hooks jobs (container and pod)

Some of the init and jobs containers need more permissions to copy/modify files. The st2actionrunner pods are most likely to have the most permissive permissions, so default to that if the deployment/job-specific securityContext values are not defined.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Development
 * Fix syntax with ensure-packs-volumes-are-writable job (#403) (by @skiedude)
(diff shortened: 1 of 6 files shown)


winem (approved):
I'd usually prefer to fallback to the least privileged securityContext but the way you implemented it is a big improvement to the current state security wise and people who will make use of it will also be skilled enough to choose are more restricted context setting. Well done!

cognifloyd:
I figure the global securityContext and podSecurityContext would have the least privileged bit. But some things like `st2web` and `st2actionrunner` and some of the init containers needed a bit more than that in my env. I figured that the init containers are somewhat opaque to the helm chart user, and I didn't want to add a lot of extra values, so I opted to have those init containers fall back to the permissions that are most likely to allow them to function.

PR #413, opened 2024-04-11
Add `image.entrypoint` value to support custom entrypoints like `dumb-init`

cognifloyd:
My actionrunner pods have exhausted the pid space due to zombie processes. So, we need something to cleanup those zombies. `dumb-init` does that. See:
- https://github.com/Yelp/dumb-init
- https://engineeringblog.yelp.com/2016/01/dumb-init-an-init-for-docker.html

Installing `dumb-init`, `pid1`, or some other entrypoint app requires installing it in the docker image (ie we have to update the docker images). But, the chart also needs to support the older images that don't include it. So, hard-coding the entry-point is not an option. Once we update the docker images to install an init process, we can update the default value in the chart to include that init process, but still allow people using an older image to set that value to `[]` and still use the chart.

Also, we would still need to include the init command in the chart somehow (not just in the docker image's `ENTRYPOINT`) because the chart overwrites the command baked into the images to pass additional config files. So, making this a value seems to be a good path forward.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -182,8 +182,9 @@ Reduce duplication of the st2.*.conf volume details
 {{- end -}}
 
 {{- define "stackstorm-ha.st2-entrypoint" -}}
-- /usr/bin/dumb-init
-- '--'
+  {{- range $.Values.image.entrypoint }}
+- {{ toYaml . }}
+  {{- end }}
 {{- end -}}
 
 # Override CMD CLI parameters passed to the startup of all pods to add support for /etc/st2/st2.secrets.conf
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -181,6 +181,11 @@ Reduce duplication of the st2.*.conf volume details
 {{- end }}
 {{- end -}}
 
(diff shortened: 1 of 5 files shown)


templates/_helpers.tpl
```diff
@@ -181,6 +181,11 @@ Reduce duplication of the st2.*.conf volume details
 {{- end }}
 {{- end -}}
 
+{{- define "stackstorm-ha.st2-entrypoint" -}}
+- /usr/bin/dumb-init
+- '--'
```
cognifloyd, line 186: I wonder about pulling this from values. That way we can change the default when the upstream docker images are adjusted. That way if someone is using an older image that doesn't have `dumb-init`, they can still use this chart.

templates/_helpers.tpl
```diff
@@ -181,6 +181,12 @@ Reduce duplication of the st2.*.conf volume details
 {{- end }}
 {{- end -}}
 
+{{- define "stackstorm-ha.st2-entrypoint" -}}
+  {{- range $.Values.image.entrypoint }}
+- {{ toYaml . }}
```
cognifloyd, line 186: The point of `toYaml` is to quote the string if needed.

winem (approved):

PR #415, opened 2024-04-12
Use rsync to copy pack contents

### LINKED ISSUE #245: Make packs/virtualenv file copy more robust

cognifloyd:
Closes #245

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -343,9 +343,14 @@ Merge packs and virtualenvs from st2 with those from st2packs images
   command:
     - 'sh'
     - '-ec'
-    - |
-      /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
-      /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+    - >
+      if command rsync; then
+        rsync -a /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
+        rsync -a /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared;
+      else
+        /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
+        /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared;
+      fi
   {{- with .securityContext | default $.Values.st2actionrunner.securityContext | default $.Values.securityContext }}
   {{/* st2actionrunner is likely the most permissive so use that if defined. */}}
   securityContext: {{- toYaml . | nindent 8 }}
@@ -365,9 +370,14 @@ Merge packs and virtualenvs from st2 with those from st2packs images
   command:
     - 'sh'
     - '-ec'
-    - |
-      /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
-      /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+    - >
+      if command rsync; then
+        rsync -a /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
+        rsync -a /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared;
+      else
+        /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
+        /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+      fi
   {{- with .Values.st2actionrunner.securityContext | default .Values.securityContext }}
   {{/* st2actionrunner is likely the most permissive so use that if defined. */}}
   securityContext: {{- toYaml . | nindent 8 }}
@@ -386,8 +396,12 @@ Merge packs and virtualenvs from st2 with those from st2packs images
   command:
     - 'sh'
     - '-ec'
-    - |
-      /bin/cp -aR /opt/stackstorm/configs/. /opt/stackstorm/configs-shared
+    - >
+      if command rsync; then
+        rsync -a /opt/stackstorm/configs/. /opt/stackstorm/configs-shared;
+      else
+        /bin/cp -aR /opt/stackstorm/configs/. /opt/stackstorm/configs-shared;
+      fi
   {{- with .Values.st2actionrunner.securityContext | default .Values.securityContext }}
   {{/* st2actionrunner is likely the most permissive so use that if defined. */}}
   securityContext: {{- toYaml . | nindent 8 }}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -344,12 +344,12 @@ Merge packs and virtualenvs from st2 with those from st2packs images
     - 'sh'
     - '-ec'
     - >
-      if hash rsync 2>/dev/null; then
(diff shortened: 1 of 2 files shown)


templates/_helpers.tpl
```diff
@@ -343,9 +343,14 @@ Merge packs and virtualenvs from st2 with those from st2packs images
...
     - '-ec'
-    - |
-      /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
-      /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+    - >
+      if command rsync; then
+        rsync -a /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
+        rsync -a /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared;
+      else
+        /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
```
cognifloyd, line 351: `-a` is equivalent to `-dR --preserve=all`:
- `-d`: no dereference (preserve links)
- `-R`: copy directories recursively
- `--preserve=all`: preserve file attributes
  - `mode`
  - `ownership` for user and group
  - `timestamps`
  - `links` for hard links
  - `context` for security context
  - `xattr` for extended attributes

Of these we need `-dR --preserve=mode` (we need mode to maintain the execute bits of action files). Depending on which user is copying the file, we might also need `--preserve=ownership`.

cognifloyd (commented):
Either we do something like this, or we make the command configurable via values.

Maybe `.utility.rsync = "rsync -rlp"`. That would mean the user would have to update this if they needed to switch to `"cp"`. Or we keep both, but allow specifying the args.
```yaml
utility:
  rsync: "rsync -a"
  cp: "cp -aR"
```

```yaml
utility:
  rsync: "rsync -rlE"
  cp: "cp -dR --preserve=mode"
```

The gotcha with making it configurable is that people could forget to include recursive, symlinks, and preserving execute bits.

Stealthii:
In an attempt to solve the ownership issue that blocks #400, we should only preserve the file attributes that matter for this purpose.

I've updated a fork of this branch with proposed changes that once verified, could be brought into this PR.

https://github.com/StackStorm/stackstorm-k8s/compare/rsync...Stealthii:stackstorm-k8s:rsync

templates/_helpers.tpl
```diff
@@ -343,9 +343,14 @@ Merge packs and virtualenvs from st2 with those from st2packs images
   command:
     - 'sh'
     - '-ec'
-    - |
-      /bin/cp -aR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
-      /bin/cp -aR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
+    - >
+      if command rsync; then
+        rsync -a /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
```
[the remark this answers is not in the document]
cognifloyd: I think I'll push a commit that switches to `-rlptD` based on https://github.com/StackStorm/stackstorm-k8s/compare/rsync...Stealthii:stackstorm-k8s:rsync

templates/_helpers.tpl
cognifloyd: I think I'll push a commit that switches to `-RP --preserve=mode,timestamps,links,xattr` based on https://github.com/StackStorm/stackstorm-k8s/compare/rsync...Stealthii:stackstorm-k8s:rsync

u1735067:
FYI, when using a pack built with Alpine 3.23.3 + rsync 3.4.1:
```
/ # command rsync; echo $?
rsync  version 3.4.1  protocol version 32
[..]
Please see the rsync(1) and rsyncd.conf(5) manpages for full documentation.
See https://rsync.samba.org/ for updates, bug reports, and answers
rsync error: syntax or usage error (code 1) at main.c(1767) [client=3.4.1]
1
```
It can be fixed with:
```
/ # command -v rsync; echo $?
/usr/bin/rsync
0
```

So this can be used:
```
if command -v rsync >/dev/null 2>&1; then
  rsync -rlptD /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
  rsync -rlptD /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared;
else
  cp -RP --preserve=mode,timestamps,links,xattr /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
  cp -RP --preserve=mode,timestamps,links,xattr /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared;
fi
```
However, it leads to:
```
rsync: [generator] failed to set times on "/opt/stackstorm/packs-shared/.": Operation not permitted (1)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.4.1]
```

Also, busybox `cp` is limited:
```
cp: unrecognized option: preserve=mode,timestamps,links,xattr
BusyBox v1.37.0 (2025-12-16 14:19:28 UTC) multi-call binary.

Usage: cp [-arPLHpfinlsTu] SOURCE DEST
or: cp [-arPLHpfinlsu] SOURCE... { -t DIRECTORY | DIRECTORY }

Copy SOURCEs to DEST

	-a	Same as -dpR
	-R,-r	Recurse
	-d,-P	Preserve symlinks (default if -R)
	-L	Follow all symlinks
	-H	Follow symlinks on command line
	-p	Preserve file attributes if possible
	-f	Overwrite
	-i	Prompt before overwrite
	-n	Don't overwrite
	-l,-s	Create (sym)links
	-T	Refuse to copy if DEST is a directory
	-t DIR	Copy all SOURCEs into DIR
	-u	Copy only newer files
```
So in the end, this is fine:
```
/bin/cp -dR /opt/stackstorm/packs/. /opt/stackstorm/packs-shared &&
/bin/cp -dR /opt/stackstorm/virtualenvs/. /opt/stackstorm/virtualenvs-shared
```

PR #417, opened 2024-05-08
Refactor unit tests to support unittests v0.4.4

jk464:
Implements #414 

Updated our `tests/unit` to support newer versions of `unittests` - for now bumping to `v0.4.4` as `v0.5.0` has a bug that impacts us (see https://github.com/helm-unittest/helm-unittest/issues/329)*

Changes required:

* `v0.3.0` of `unittests` adds `jsonPath` which required updating our `path` references to use a `jq` style compatible syntax
```diff
@@ -57,55 +57,55 @@ tests:
       # So, we use isNotNull instead.
       # see: https://github.com/quintush/helm-unittest/issues/122
       - isNotNull:
-          path: metadata.labels.[app.kubernetes.io/name]
+          path: metadata.labels["app.kubernetes.io/name"]
       - isNotNull:
```
* `isNull` changed behavior, such that if the path _exists_ but has an "empty" value - the assertion now fails, i.e. the below `k8s` definition _doesn't_ pass (when it did before), since having empty `keys` doesn't do anything functionally when applied to `k8s`, and in my opinion atleast, looks displeasing/confusing - I fixed up the `helm` templates to not produce empty `imagePullSecrets` if nothing is being set - fixing the test again.
```diff
@@ -38,8 +38,8 @@ spec:
           {{- toYaml .Values.st2auth.annotations | nindent 8 }}
         {{- end }}
     spec:
-      imagePullSecrets:
       {{- if .Values.image.pullSecret }}
+      imagePullSecrets:
       - name: {{ .Values.image.pullSecret }}
       {{- end }}
       initContainers:
...
```
* Some assertions have been renamed as of `v0.3.1` (including `isNull`, which seems to have also been when the aforementioned behavior change was introduced), other that the above issue, nothing else behaviorally seems to have changed and the old names still work - however to ensure we're future proofed in-case they are removed (in the docs atleast `isNull` and `isNotNull` are mentioned to be "deprecated", I've update the following assertions:
    * `isNotNull` with `exists`
    * `isNull` with `notExists`
    * `isEmpty` with `isNullOrEmpty`
    * `isNotEmpty` with `isNotNullOrEmpty`

With those changes, our tests now work on `v0.4.4` (and should work on `v0.5.x` once the previously mentioned bug is fixed)

The pipeline also has been updated to use `v0.4.4`

----------------------------------------------------
* Note however, testing with the broken test confirms that our tests are otherwise compatible with `v0.5.0` so once we have a fixed `v0.5.x` version we'll be able to bump to that with hopefully no further changes.

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -35,7 +35,7 @@ jobs:
         # We should periodically check to see if another fork has taken over maintenance,
         # as the de-facto "best" fork has changed several times over the years.
         run: |
-          helm plugin install https://github.com/quintush/helm-unittest --version v0.2.11
+          helm plugin install https://github.com/helm-unittest/helm-unittest.git  --version v0.4.4
 
       - name: Install chart dependencies
         run: |
@@ -44,4 +44,4 @@ jobs:
       - name: Run helm-unittest
         # by default looks for tests/*_test.yaml
         run: |
-          helm unittest --color --helm3 -f 'tests/unit/*_test.yaml' .
+          helm unittest --color -f 'tests/unit/*_test.yaml' .
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Development
+* Updated our tests/unit to support newer versions of `unittests` - for now bumping to `v0.4.4` as `v0.5.0` has a bug that impacts us (see helm-unittest/helm-unittest#329), but testing around the bug shows `v0.5.x` should also "just work" (#414) (by @jk464)
 
 ## v1.1.0
 * Fix syntax with ensure-packs-volumes-are-writable job (#403, #411) (by @skiedude)
--- a/templates/deployments.yaml
+++ b/templates/deployments.yaml
@@ -38,8 +38,8 @@ spec:
           {{- toYaml .Values.st2auth.annotations | nindent 8 }}
         {{- end }}
     spec:
-      imagePullSecrets:
       {{- if .Values.image.pullSecret }}
+      imagePullSecrets:
       - name: {{ .Values.image.pullSecret }}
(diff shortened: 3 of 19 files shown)


.github/workflows/unit.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
         # We should periodically check to see if another fork has taken over maintenance,
         # as the de-facto "best" fork has changed several times over the years.
         run: |
-          helm plugin install https://github.com/quintush/helm-unittest --version v0.2.11
+          helm plugin install https://github.com/helm-unittest/helm-unittest.git  --version v0.4.4
```
cognifloyd, line 38: Is `.git` required?
```suggestion
          helm plugin install https://github.com/helm-unittest/helm-unittest --version v0.4.4
```

cognifloyd (changes requested):
Would you please update the quintush/helm-unittest references in `tests/README.md` as well?

And I have one non-blocking nitpick in `.github/workflows/unit.yaml`.

Everything else looks great! Thank you for picking this up!

jk464:
> Would you please update the quintush/helm-unittest references in `tests/README.md` as well?
> 
> And I have one non-blocking nitpick in `.github/workflows/unit.yaml`.
> 
> Everything else looks great! Thank you for picking this up!

Documentation updated

.github/workflows/unit.yaml
jk464: `.git` doesn't seem to be required (I tested and it works w/o) - but the `unittest` documentation https://github.com/helm-unittest/helm-unittest/tree/main?tab=readme-ov-file#install has `.git` in the install command.

Unsure if we want to remove it, or leave it as is since thats the documented install FQDN.

cognifloyd (approved):

Issue #433, opened 2025-09-09
Update external dependencies


akshat-rubrik:
Bitnami is moving image locations: https://github.com/bitnami/containers/issues/83267
we will have to update default path of external dependencies to the new location.

ottacom:
... it's quite evident that stackstorm is not really updated

fdrab:
perhaps this is not the most important update to merge right now. it's just a failing test, you can just update your own values.yaml and it'll download the correct dependencies.

long term though, this needs to change if this chart is supposed to deploy the dependent packages. I have switched to deploying the official images from vendors (not via helm though)

ottacom:
I'm not getting sorry...
The version of mongodb for instance can't be > 4.4 from bitnami.
Bitnami removed all the previous version  and despite I tried to use a regular version of mongodb  4.4 (not bitnami branded) nothing seems working , same for rabbit-mq and redis.
This tool is intended to be used for automation in a large scale, and it's actually the brain of the infrastructure, can't be outdated.
I tried the docker version that is AFAIK not supposed to be used for prod.
The packaged version is supporting only ubuntu 2020 , which is alomst EOS .
Maybe I'm missing something but if you have some suggestion to install Stackstorm ok K8s pls let me know , I'll be so glad to use it .
Thanks a lot

winem:
Preferably, we switch to sources for the helm charts that will still be maintained. We should check if we can just switch to external-dns from the k8s sig repository: https://kubernetes-sigs.github.io/external-dns/

fdrab:
> Preferably, we switch to sources for the helm charts that will still be maintained. We should check if we can just switch to external-dns from the k8s sig repository: https://kubernetes-sigs.github.io/external-dns/

For External-DNS, why not. Are there maintained equivalents for MongoDB / REDIS / RabbitMQ?

guzzijones:
working on this now

guzzijones:
see #440

PR #435, opened 2025-10-01
added bitnamilegacy overrides (#433)

fdrab:
Added overrides for repository locations as they were migrated to bitnamilegacy and are no longer maintained. This is a temporary workaround until a better solution is found.


winem:
Hi, 

I think we can drop bitnami at least for the external-dns by switching to https://kubernetes-sigs.github.io/external-dns/ 

Since it's not super critical to get this PR merged as soon as possible I'd love to see the change as part of this PR. 

Please let me know if you can't do it or prefer to do it in a dedicated PR for any reason. In that case I'd be happy to support you.

fdrab:
3rd mongo replica failed to start, for whatever reason. worked fine on my PC. no clue how to restart it

fdrab:
closing as guzzijones added non-bitnami implementations for dependencies instead of relying on bitnami-legacy

PR #440, opened 2026-03-20
Deps fix

guzzijones:
1. change to rabbitmq operator dependency
2. change to mongocommunity operator dependency
3. change to valkey helm dependency
4. add missing rbac standard groups initiation via python script job.  system_admin and observer


guzzijones:
@cognifloyd any initial feedback before i go down the github ci/cd rabbithole?

guzzijones:
is there a way to proxy dependency of dependency helm charts?  i have a wrapper helm chart and i could not get this to work offline without commiting the dependent charts.

guzzijones:
@nzlosh are the github actions disabled here?  They are not running and the checks say pending. Thanks.

